PR #5, opened 2016-12-20
Use formatting mode

chentao1596:
I think we should use formatting mode rather than plus, do you think so? thank you!


chentao1596:
@bowei @MrHohn hi, cc, thank you!

bowei:
can you fix your commit message to be more descriptive:

"Use string format instead of append"

chentao1596:
@bowei Hi, i've  changed it, PTAL, thank you!

PR #11, opened 2016-12-28
Port over sidecar e2e to go, make more robust to flakes

bowei:


cmd/sidecar-e2e/main.go
```diff
@@ -0,0 +1,353 @@
...
+}
+
+func (t *test) dump() {
+	t.lock.Lock()
+	defer t.lock.Unlock()
+
+	fmt.Println("BEGIN dnsmasq ====")
+	fmt.Println(t.dnsmasqOutput)
+	fmt.Println("END dnsmasq ====")
+	fmt.Println()
```
MrHohn: Minor: may replace this with `\n`?

cmd/sidecar-e2e/main.go
```diff
@@ -0,0 +1,353 @@
...
+
+	defer response.Body.Close()
+	buf, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		log.Fatal(err)
+	}
+	t.metricsOutput = string(buf)
+}
+
+func (t *test) dump() {
```
MrHohn: Why dump uses fmt.Println() but not log?

bowei: println outputs to stdout which is captured by the harness, log uses stderr which is shown to the user

PR #16, opened 2016-12-31
Fix some go style mistakes about fmt.Errorf

chentao1596:
I think we should follow this request: https://github.com/golang/go/wiki/CodeReviewComments#error-strings, thank you!


chentao1596:
@bowei @MrHohn hi, cc, thank you!

PR #24, opened 2017-01-10
Adds DNS-schema specification

bowei:
Original patch: https://github.com/kubernetes/kubernetes/pull/38408

Add a DNS-schema spec to create a baseline for any DNS-based
service discovery component to follow.

Rewrite describing records that must exist

Re-wrote the specification to indicate the records that must exist.
This change the behavior of a few queries that were acting something
like wildcards. For example, querying for an SRV record with the
service name and no _port._proto prefix is no longer required to
return any records.

Other minor updates to link to external references and correct typos.

Fix formatting of headless SRV answer example.

Add reference to rcode

Add numbering, cleanup non-ASCII characters, address deprecation warning

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -1,0 +1,185 @@
+# Kubernetes DNS-Based Service Discovery
+
+## 0 - About This Document
+
+This document is a specification for DNS-based Kubernetes service
+discovery. While service discovery in Kubernetes may be provided via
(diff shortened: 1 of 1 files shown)


johnbelamaric:
LGTM

smarterclayton:
LGTM

PR #25, opened 2017-01-11
Reverse DNS records for named headless services

sadlil:
Add reverse DNS records for pods' IPs to named headless services' with Pod Hosts FQDN (`<podHostName>.<serviceName>.<namespace>.svc.cluster.local`).

Original PR: kubernetes/kubernetes#38376

Resolves Issues
 - kubernetes/kubernetes#26752
 - kubernetes/kubernetes#33470

Spec:
**Records for a Named Headless Service**
Type: PTR
f a ready endpoint exists with IP equal to the IP of exists, and the endpoint has a non-empty hostname with value <hostname>, then a PTR record with <hostname>.<service>.<ns>.svc.<zone> will be returned.

@thockin @bowei @smarterclayton @harryge00

### DIFF
--- a/pkg/dns/dns.go
+++ b/pkg/dns/dns.go
@@ -242,13 +242,10 @@ func (kd *KubeDNS) setEndpointsStore() {
 		&v1.Endpoints{},
 		resyncPeriod,
 		kcache.ResourceEventHandlerFuncs{
-			AddFunc: kd.handleEndpointAdd,
-			UpdateFunc: func(oldObj, newObj interface{}) {
-				// TODO: Avoid unwanted updates.
-				kd.handleEndpointAdd(newObj)
-			},
-			// No DeleteFunc for EndpointsStore because endpoint object will be deleted
-			// when corresponding service is deleted.
+			AddFunc:    kd.handleEndpointAdd,
+			UpdateFunc: kd.handleEndpointUpdate,
+			// If Service is named headless need to remove the reverse dns entries.
+			DeleteFunc: kd.handleEndpointDelete,
 		},
 	)
 }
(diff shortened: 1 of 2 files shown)


thockin:
There is now a specification in this repo - can you also update that, please?

sadlil:
@thockin i think [2.4.3](https://github.com/kubernetes/dns/blob/master/docs/specification.md#243---ptr-records) already says what this PR does.

pkg/dns/dns.go
```diff
@@ -323,6 +320,76 @@ func (kd *KubeDNS) handleEndpointAdd(obj interface{}) {
 	}
 }
 
+func (kd *KubeDNS) handleEndpointUpdate(oldObj, newObj interface{}) {
+	if o, ok := oldObj.(*v1.Endpoints); ok {
```
bowei: short circuit these and log error and return if the type cast does not work. For longer scoped variables, I prefer more than one letter names.

```go
func (kd *KubeDNS) handleEndpointUpdate(oldObj, newObj interface{}) {
  oldEndpoints, ok ;= oldObj.(*v1...)
  newEndpoints, ok := newObj.(*v1...)
  if oldEndpoints == nil { 
    glog.Errorf("oldObj was not of type v1.Endpoints)
    // etc
  }
// etc

File: pkg/dns/dns.go [hunk 2]
```diff
@@ -323,6 +320,76 @@ func (kd *KubeDNS) handleEndpointAdd(obj interface{}) {
...
+			}
+
+			// TODO: Avoid unwanted updates.
+			kd.handleEndpointAdd(newObj)
+		}
+	}
+}
+
+func (kd *KubeDNS) handleEndpointDelete(obj interface{}) {
+	if e, ok := obj.(*v1.Endpoints); ok {
```
bowei: short circuit log error and return (see comment above)

pkg/dns/dns.go
```diff
@@ -323,6 +320,76 @@ func (kd *KubeDNS) handleEndpointAdd(obj interface{}) {
...
+		svc, err := kd.getServiceFromEndpoints(e)
+		if svc != nil && err == nil {
+			if !v1.IsServiceIPSet(svc) {
+				// When endpoints for Named headless services deleted, delete old reverse dns records.
+				for idx := range e.Subsets {
+					for subIdx := range e.Subsets[idx].Addresses {
+						address := &e.Subsets[idx].Addresses[subIdx]
+						endpointIP := address.IP
+						if _, has := getHostname(address); has {
+							delete(kd.reverseRecordMap, endpointIP)
```
bowei: Do you need to grab the lock to do this update?

pkg/dns/dns.go
```diff
@@ -323,6 +320,76 @@ func (kd *KubeDNS) handleEndpointAdd(obj interface{}) {
...
+								}
+							}
+						}
+					}
+
+					// Remove all old PTR records for the endpoint that are not
+					// in new endpoint, or
+					// the addresses that are no longer named.
+					for k := range oldAddressMap {
+						delete(kd.reverseRecordMap, k)
```
bowei: I think the mutex needs to be held for the deletion

sadlil:
@bowei updated with locking and short circuit.

reAsOn2010:
@sadlil I used this PR for my HBase deployment, which is strongly relied on reverse DNS lookup. It worked as what I wanted. Hoping for merge and many thanks~

thockin:
Make sure this gets a release note when it integrates to main repo.

bowei:
@sadlil -- you will have to open a PR integrating kube-dns:1.11.0. Remember the release note.

sadlil:
@bowei  opened kubernetes/kubernetes#39925.

pkg/dns/dns.go
```diff
@@ -323,6 +320,94 @@ func (kd *KubeDNS) handleEndpointAdd(obj interface{}) {
...
+	}
+
+	// TODO: Avoid unwanted updates.
+	kd.handleEndpointAdd(newObj)
+}
+
+func (kd *KubeDNS) handleEndpointDelete(obj interface{}) {
+	endpoints, ok := obj.(*v1.Endpoints)
+	if !ok {
+		glog.Errorf("obj type assertion failed! Expected 'v1.Endpoints', got %T", obj)
```
kargakis: You can get a tombstone which you could cast again and get the endpoint that was deleted out of it. See the ds delete resource handler for example: https://github.com/kubernetes/kubernetes/blob/f7305e6f4375a34cb6e2f47b6e164527370e8871/pkg/controller/daemon/daemoncontroller.go#L174

sadlil: The DNS package didn't use the tombstone [pattern.](https://github.com/kubernetes/dns/blob/master/pkg/dns/dns.go#L254) I tried to followed what was done here.
If you like i can send a pr implementing the tomb stone pattern.

kargakis: @bowei @thockin is there any specific reason not to use tombstones?

smarterclayton:
How does this handle multiple endpoints all reporting the same IP with different hostnames?  I did not see that in the code, but multiple services can represent the same pod, and there's nothing that prevents a caller from setting their own endpoints that have hostnames and use the pod.  For example, a controller that copied endpoints to identify which pod is the "master" could easily have the hostname and IP the same as another service, and thus result in ordering of updates changing what is in the cache.

thockin:
Oh snap!  THIS was why we didn't do it in the first place - it requires us to confirm with the pod, and we don't want DNS watching every Pod.

We can really only have one reverse record per IP.  Without checking that this is the same Service that pod elected as Subdomain, we can't do this safely.

Do we need to revert this?  Sad because it will work sometimes...

sadlil:
@thockin, @smarterclayton this problem can be avoided if we do a get Pod only when `address.Hostname` is set, and check for the `Pod.Spec.Subdomain == svc.Name`. Is this reasonable?
I can have a PR ready for this.

sadlil:
@thockin @smarterclayton i have created a sample PTR with `address.Hostname` is set, and check for the `Pod.Spec.Subdomain == svc.Name` solution.
To further reduce cost we may enforce a annotation over Headless service. `service.beta.kubernetes.io/ptr: true`. PTR records will only be generated if the annotation exists.

Let me know what you think.

thockin:
I am back from vacation and scrubbing email.

Did we do anything with this?  This is a real issue which makes PTR records
almost useless.  I'm going to open a new issue.

chrislovecnm:
We are running into a nasty PTR issue.

Create package let's call it alpha. It will create pods named alpha-0.alpha.default.cluster.local. The default naming for a headless service and statefulset.

1. Use a statefulset on hostnetwork, and nodeport
2. Use a headless service with said set
3. Use a nodeport service with set set

Basically one pod on one node all running on 9042. Dns is both in nodeport and headless services.

Now uninstall alpha.  Update the manifest and rename the stateful set as beta. Install beta.

Now do reverse ip lookup, the PTR record is set to alpha-0 old pod name. Lions tigers and bears of my. This was a nasty one ;(

Basically with the manifest that I am running  with 1.7.4 PTR records are getting set once and then never updated.

thockin:
Yeah, I'm afraid PTR as implemented is pretty broken.  I'm not sure what the right fix is.  The proposed fix is pretty gnarly...

PR #28, opened 2017-01-13
Adding an option to passthrough upstream nameservers to skydns.

corlettb:
Currently the upstream nameservers are hardcoded to be taken from
/etc/resolv.conf.

Allowing this option to be passed in can offer greater flexibility
for varying setups where kube-dns is not run within a container.

### DIFF
--- a/cmd/kube-dns/app/options/options.go
+++ b/cmd/kube-dns/app/options/options.go
@@ -49,6 +49,8 @@ type KubeDNSConfig struct {
 
 	ConfigDir    string
 	ConfigPeriod time.Duration
+
+	NameServers string
 }
 
 func NewKubeDNSConfig() *KubeDNSConfig {
@@ -66,6 +68,8 @@ func NewKubeDNSConfig() *KubeDNSConfig {
 
 		ConfigPeriod: 10 * time.Second,
 		ConfigDir:    "",
+
+		NameServers: "",
 	}
 }
 
@@ -143,6 +147,11 @@ func (fv federationsVar) Type() string {
 func (s *KubeDNSConfig) AddFlags(fs *pflag.FlagSet) {
 	fs.Var(clusterDomainVar{&s.ClusterDomain}, "domain",
 		"domain under which to create names")
+
+	fs.StringVar(&s.NameServers, "nameservers", s.NameServers,
+		"List of ip:port, separated by commas of nameservers to forward queries to. "+
+			"If set, overrides upstream servers taken from the nameserver option in /etc/resolv.conf. "+
+			"Example: 8.8.8.8:53,8.8.4.4 (default port is 53)")
 
 	fs.StringVar(&s.KubeConfigFile, "kubecfg-file", s.KubeConfigFile,
 		"Location of kubecfg file for access to kubernetes master service;"+
--- a/cmd/kube-dns/app/server.go
+++ b/cmd/kube-dns/app/server.go
@@ -18,9 +18,13 @@ package app
 
 import (
 	"fmt"
+	"net"
 	"net/http"
 	"os"
 	"os/signal"
+	"regexp"
+	"strconv"
+	"strings"
 	"syscall"
 
 	"github.com/golang/glog"
@@ -43,6 +47,7 @@ type KubeDNSServer struct {
 	healthzPort    int
 	dnsBindAddress string
 	dnsPort        int
+	nameServers    string
 	kd             *dns.KubeDNS
 }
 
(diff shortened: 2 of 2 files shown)


cmluciano:
Do you envision that these are nameservers that would not be present on the host? Any reason for these not to be added to the host instead?

bowei:
@corlettb -- Can you describe your deployment scenario where kube-dns is not run inside a container? It seems somewhat out of the norm of the environment in which we expect to run kube-dns.

corlettb:
We run in a mixed environment, some apps inside kubernetes and some apps outside. 

Our routing is configured such that service ips work from outside kubernetes along with kube-dns entries with a delegated zone. Currently we have kube-dns running inside kubernetes and use the resolv.conf override to set its upstream resolvers. This works as you'd expect.

When we did our first iteration of kubernetes we had most kubernetes components running inside kubenetes/docker. Sadly we had a number of issues. We found running k8s components inside itself made it a very difficult thing to troubleshoot. Especially for things like dns lookup hangs. This is not aided by busybox containers missing a lot of the tools required to troubleshoot.

There is now a push to move all core kubernetes components to run outside of itself. I know you could probably argue kube-dns isn't a core component. We've had that discussion ourselves.

The reason you need the nameserver override in that situation is because you may end up with infinite loop lookups if you want dns to operate in a normal way on the machine running kube-dns. Kube-dns can still end up forwarding requests for its zone upstream if it doesn't know about them.

machine -> dns server [zone delegation] -> kube-dns [doesn't have an entry for dns] -> dns server [zone delegation] -> kube-dns ........

I did consider writing a startup script that would modify /etc/resolv.conf, start kube-dns, sleep 5 seconds and then revert /etc/resolv.conf. Not the best solution.

cmd/kube-dns/app/options/options.go
```diff
@@ -138,6 +140,9 @@ func (s *KubeDNSConfig) AddFlags(fs *pflag.FlagSet) {
 	fs.Var(clusterDomainVar{&s.ClusterDomain}, "domain",
 		"domain under which to create names")
 
+	fs.StringVar(&s.NameServers, "nameservers", s.NameServers,
+		"List of ip:port, separated by commas of recursive nameservers to forward queries to.")
```
bowei: The `:port` is optional, is it not?

Remove `recursive` (upstream may not necessarily be such)

Add "If set, overrides upstream servers taken from the `nameserver` option in `/etc/resolv.conf`.

Give an example of a valid value in the description: "Example: 8.8.8.8,.8.8.4.4"

cmd/kube-dns/app/server.go
```diff
@@ -103,6 +106,9 @@ func (server *KubeDNSServer) Run() {
 	server.setupHandlers()
 
 	glog.V(0).Infof("Status HTTP port %v", server.healthzPort)
+	if server.nameServers != "" {
+		glog.V(0).Infof("Upstream namesservers: %s", server.nameServers)
```
bowei: name**s**server (spelling)

cmluciano:
May be related https://github.com/docker/docker/pull/22408

corlettb:
I've updated the pull request with @bowei 's recommendations in mind. I've also added a default port and validation.

cmd/kube-dns/app/server.go
```diff
@@ -142,6 +148,11 @@ func (d *KubeDNSServer) startSkyDNSServer() {
 		Domain:  d.domain,
 		DnsAddr: fmt.Sprintf("%s:%d", d.dnsBindAddress, d.dnsPort),
 	}
+	if d.nameServers != "" {
+		for _, hostPort := range strings.Split(d.nameServers, ",") {
```
bowei: did you fix this? hostPort does not seem like the correct name for the variable

corlettb: I took the variable name from the skydns code. I guess hostPort means a host and a port. Happy to change the name though. I'll update the pull request tomorrow morning.

bowei: ah -- that's where it is from, somewhat confusing

bowei:
looks like there are conflicts, otherwise looks good

bowei:
@corlettb  -- also would be great to squash all of the commits together into one

corlettb:
Changes merged and squashed.

bowei:
Travis seems to not be picking this up -- can you push the branch again?

dalehamel:
This solves a problem that we are running into where we need to override the nameservers skydns is talking to.

I'm curious what the schedule would be to get this functionality included into a release, with a container built ?

bowei:
It should make 1.6

bowei:
@dalehamel @corlettb 

Pushed:
```
gcr.io/google_containers/k8s-dns-kube-dns-amd64:1.13.0
gcr.io/google_containers/k8s-dns-sidecar-amd64:1.13.0
gcr.io/google_containers/k8s-dns-dnsmasq-amd64:1.13.0
...
```

You will need to update the main repository with the new image to have the changes be part of the release. Here is an example PR that does something similar: https://github.com/kubernetes/kubernetes/pull/40382

liggitt:
would you like me to update my PR to 1.13.0?

bowei:
@liggitt -- which ever is easier for you

liggitt:
I'll actually leave kubernetes/kubernetes#40382 as-is since it's got green tests

PR #30, opened 2017-01-13
WIP: Make manifest lists of the dns images

luxas:
... and base everything on busybox to minimize the delta between amd64 and non-amd64 images

Also I tried to make the Makefiles simpler, let me know what you think @bowei and others!

A manifest list is an image that's kind of a placeholder for other arch-specific images, so basically if we have `k8s-dns-dnsmasq:version` in a manifest and are using docker; docker will fetch the right `-arch` version for us automatically!

Tested it locally, and it seems like I got it working. I also added s390x to the list

### DIFF
--- a/Dockerfile.kube-dns
+++ b/Dockerfile.kube-dns
@@ -14,7 +14,6 @@
 
 FROM ARG_FROM
 
-MAINTAINER Bowei Du <bowei@google.com>
-
 ADD bin/ARG_ARCH/ARG_BIN /ARG_BIN
+USER nobody:nogroup
 ENTRYPOINT ["/ARG_BIN"]
--- a/images/dnsmasq/Makefile
+++ b/images/dnsmasq/Makefile
@@ -18,44 +18,37 @@ ARCH ?= amd64
 DNSMASQ_VERSION ?= dnsmasq-2.76
 CONTAINER_PREFIX ?= k8s-dns
 
-ALL_ARCH := amd64 arm arm64 ppc64le
-IMAGE := $(CONTAINER_PREFIX)-dnsmasq-$(ARCH)
-COMPILE_IMAGE := gcr.io/google_containers/kube-cross:v1.6.2-2
+ALL_ARCH := amd64 arm arm64 ppc64le s390x
+IMAGENAME := $(CONTAINER_PREFIX)-dnsmasq
+# TODO: Use gcr.io/google_containers/kube-cross:v1.7.4-2
+COMPILE_IMAGE := gcr.io/google_containers/kube-cross:v1.7.4-1
 OUTPUT_DIR := _output/$(ARCH)
+ML_PLATFORMS=linux/amd64,linux/arm,linux/arm64,linux/ppc64le,linux/s390x
 
 ifeq ($(ARCH),amd64)
-	BASEIMAGE ?= alpine:3.4
-	COMPILE_IMAGE := alpine:3.4
+	BASEIMAGE ?= busybox:glibc
+	CC ?= gcc
 else ifeq ($(ARCH),arm)
+	# TODO: Switch to using armhf/ and arm-linux-gnueabihf
 	BASEIMAGE ?= armel/busybox:glibc
-	TRIPLE    ?= arm-linux-gnueabi
-	QEMUARCH  := arm
+	CC        ?= arm-linux-gnueabi-gcc
 else ifeq ($(ARCH),arm64)
 	BASEIMAGE ?= aarch64/busybox:glibc
-	TRIPLE    ?= aarch64-linux-gnu
-	QEMUARCH  := aarch64
+	CC        ?= aarch64-linux-gnu-gcc
 else ifeq ($(ARCH),ppc64le)
 	BASEIMAGE ?= ppc64le/busybox:glibc
-	TRIPLE    ?= powerpc64le-linux-gnu
-	QEMUARCH  := ppc64le
+	CC        ?= powerpc64le-linux-gnu-gcc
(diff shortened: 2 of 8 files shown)


luxas:
cc @MrHohn also

luxas:
ping @bowei @MrHohn please take a look as soon as possible

images/dnsmasq/Makefile
```diff
@@ -89,74 +82,56 @@ all-containers: $(addprefix containers-, $(ALL_ARCH))
...
 
-$(PUSH_STAMP): $(CONTAINER_STAMP)
-	@echo "pushing  :" $(REGISTRY)/$(IMAGE):$(VERSION)
-	@$(DOCKER) push $(REGISTRY)/$(IMAGE):$(VERSION)
+$(PUSH_STAMP): $(CONTAINER_STAMP) gcr-login
+	@echo "pushing  :" $(REGISTRY)/$(IMAGENAME)-$(ARCH):$(VERSION)
+	@docker push $(REGISTRY)/$(IMAGENAME)-$(ARCH):$(VERSION)
 	@cat $< > $@
 
 .PHONY: clean
```
bowei: add manifest-tool to clean and any other generated file (I think .emptydir is created as well)

images/dnsmasq/Makefile
```diff
@@ -89,74 +82,56 @@ all-containers: $(addprefix containers-, $(ALL_ARCH))
...
 test: containers
-	@ARCH=$(ARCH) IMAGE=$(REGISTRY)/$(IMAGE):$(VERSION) ./validate.sh
+	@ARCH=$(ARCH) IMAGE=$(REGISTRY)/$(IMAGENAME)-$(ARCH):$(VERSION) ./validate.sh
+
+./manifest-tool:
+	curl -sSL https://github.com/luxas/manifest-tool/releases/download/v0.3.0/manifest-tool > manifest-tool
+	chmod +x manifest-tool
 
+gcr-login:
+ifeq ($(findstring gcr.io,$(PREFIX)),gcr.io)
```
bowei: This needs to be REGISTRY

images/dnsmasq/Makefile
```diff
@@ -89,74 +82,56 @@ all-containers: $(addprefix containers-, $(ALL_ARCH))
...
-		-q -t $(REGISTRY)/$(IMAGE):$(VERSION) $(OUTPUT_DIR) > $@
+		-q -t $(REGISTRY)/$(IMAGENAME)-$(ARCH):$(VERSION) $(OUTPUT_DIR) > $@
 
 .PHONY: test
 test: containers
-	@ARCH=$(ARCH) IMAGE=$(REGISTRY)/$(IMAGE):$(VERSION) ./validate.sh
+	@ARCH=$(ARCH) IMAGE=$(REGISTRY)/$(IMAGENAME)-$(ARCH):$(VERSION) ./validate.sh
+
+./manifest-tool:
+	curl -sSL https://github.com/luxas/manifest-tool/releases/download/v0.3.0/manifest-tool > manifest-tool
```
bowei: What does manifest-tool do? Is it essential for the build process. I would prefer to minimize external dependencies (e.g. things that need to be downloaded) for the build.

Dockerfile.kube-dns
```diff
@@ -14,7 +14,6 @@
 
 FROM ARG_FROM
 
-MAINTAINER Bowei Du <bowei@google.com>
-
 ADD bin/ARG_ARCH/ARG_BIN /ARG_BIN
+USER nobody:nogroup
```
thockin: will this work?  DNS wants to bind to :53 which is < 1024

thockin:
I would love an equivalent changeset to https://github.com/thockin/go-build-template :)

Dockerfile.kube-dns
luxas: oh, didn't think about that, just harmonized the Dockerfiles automatically, will fix

images/dnsmasq/Makefile
cmluciano: Looks like this is a library that creates multi-os [docker manifest files](https://github.com/docker/distribution/blob/master/docs/spec/manifest-v2-2.md).

images/dnsmasq/Makefile
```diff
@@ -18,44 +18,37 @@ ARCH ?= amd64
...
+IMAGENAME := $(CONTAINER_PREFIX)-dnsmasq
+# TODO: Use gcr.io/google_containers/kube-cross:v1.7.4-2
+COMPILE_IMAGE := gcr.io/google_containers/kube-cross:v1.7.4-1
 OUTPUT_DIR := _output/$(ARCH)
+ML_PLATFORMS=linux/amd64,linux/arm,linux/arm64,linux/ppc64le,linux/s390x
 
 ifeq ($(ARCH),amd64)
-	BASEIMAGE ?= alpine:3.4
-	COMPILE_IMAGE := alpine:3.4
+	BASEIMAGE ?= busybox:glibc
```
cmluciano: Does this build correctly? I believe we add several packages with apk. Is apk present in busybox:glibc?

cmluciano:
Is this still valid?

luxas:
@cmluciano I wish it was merged long time ago, but it's still pending due to that gcr.io doesn't support v2 schema 2 (-> no manifest lists can be pushed) :(

luxas:
/remove-lifecycle stale

@mkumatag is looking into this now that we're unblocked...

luxas:
Closing this as it's out of date and we're targeting that CoreDNS will get this functionality instead

PR #31, opened 2017-01-14
Fixed sidecar running group for arm arch

GheRivero:
The base image used to build the containers for the arm arch
(both arm and arm64) uses the pair nobody:nogroup while amd64
arch uses nobody:nobody

### DIFF
--- a/Dockerfile.sidecar
+++ b/Dockerfile.sidecar
@@ -18,5 +18,5 @@ MAINTAINER Bowei Du <bowei@google.com>
 
 ADD bin/ARG_ARCH/ARG_BIN /ARG_BIN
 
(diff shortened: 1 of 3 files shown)


bowei:
Does arm/arm64 not have a `nobody` user? I would like to know what effect this change has. Does the daemon no longer run? I tried the image on a Debian-based ARM platform and it run correctly.

GheRivero:
What the base image kube-dns is based on doesn't have is the 'nobody' group. It's called 'nogroup'. When running the container, it errors with: "start returned error: linux spec user: unable to find group nobody: no matching entries in group file". This change doesn't effect any of the current behavior. Only change the group for the ARM platforms

PR #36, opened 2017-01-23
Godep

bowei:
Adds a wrapper script for dealing with godep


docs/go-dependencies.md
```diff
@@ -0,0 +1,66 @@
...
+save [PKG1 PKG2 ...]
+  save PKGs as new dependencies. Note: the PKGs must be referenced
+  from the code, otherwise godep will ignore the new packages.
+
+-h|--help
+  This help message.
+
+USAGE
+
+Add a new package:
```
MrHohn: Why adding a new package needs not be run in container, but updating a package need to be in container?

docs/go-dependencies.md
```diff
@@ -0,0 +1,66 @@
...
+
+  (It is recommended to do this manually due to the fragility of godep update)
+
+  # Enter the container interactively.
+  $ build/dep.sh enter -u
+
+  # inside container
+  $ rm -rf /src/$DEP # repo root
+  $ go get $DEP/...
+  # Change code in Kubernetes, if necessary.
```
MrHohn: What does this step do?

docs/go-dependencies.md
bowei: It is run in a container. The `save` command wraps the process.

docs/go-dependencies.md
```diff
@@ -0,0 +1,66 @@
...
+  $ rm -rf Godeps
+  $ rm -rf vendor
+  $ ./build/dep.sh save
+  $ exit
+
+  $ git checkout -- $(git status -s | grep "^ D" | awk '{print $2}' | grep ^Godeps)
+
+Verifying Godeps.json match:
+
+  $ build/dep/sh verify
```
bowei: done

docs/go-dependencies.md
bowei: Downloads a new copy of $DEP

MrHohn: So the `Change code in Kubernetes` step is also done by `go get $DEP/...`?

bowei: Note: the PKGs must be referenced
 +  from the code, otherwise godep will ignore the new packages.

MrHohn: Thanks for explaining. Could you make this file in sync as the usage in dep.sh?

PR #39, opened 2017-01-24
Add support for pulling config from a directory

liggitt:
now that optional configmaps are possible, kube-dns should support pulling config from a mounted directory

this PR refactors the existing API-based configmap source and adds a directory-based source

c.f.
https://github.com/kubernetes/kubernetes/pull/36775#issuecomment-261412030
https://github.com/kubernetes/kubernetes/pull/38816
https://github.com/kubernetes/kubernetes/pull/39981

### DIFF
--- a/cmd/kube-dns/app/options/options.go
+++ b/cmd/kube-dns/app/options/options.go
@@ -46,6 +46,9 @@ type KubeDNSConfig struct {
 
 	ConfigMapNs string
 	ConfigMap   string
+
+	ConfigDir    string
+	ConfigPeriod time.Duration
 }
 
 func NewKubeDNSConfig() *KubeDNSConfig {
@@ -60,6 +63,9 @@ func NewKubeDNSConfig() *KubeDNSConfig {
 
 		ConfigMapNs: api.NamespaceSystem,
 		ConfigMap:   "", // default to using command line flags
+
+		ConfigPeriod: 10 * time.Second,
+		ConfigDir:    "",
 	}
 }
 
@@ -155,15 +161,21 @@ func (s *KubeDNSConfig) AddFlags(fs *pflag.FlagSet) {
 		"a comma separated list of the federation names and their corresponding"+
 			" domain names to which this cluster belongs. Example:"+
 			" \"myfederation1=example.com,myfederation2=example2.com,myfederation3=example.com\"."+
-			" It is an error to set both the federations and config-map flags.")
-	fs.MarkDeprecated("federations", "use config-map instead. Will be removed in future version")
+			" It is an error to set both the federations and config-map or config-dir flags.")
+	fs.MarkDeprecated("federations", "use config-dir instead. Will be removed in future version")
 
 	fs.StringVar(&s.ConfigMapNs, "config-map-namespace", s.ConfigMapNs,
 		"namespace for the config-map")
 	fs.StringVar(&s.ConfigMap, "config-map", s.ConfigMap,
 		"config-map name. If empty, then the config-map will not used. Cannot be "+
-			" used in conjunction with federations flag. config-map contains "+
+			"used in conjunction with federations or config-dir flag. config-map contains "+
 			"dynamically adjustable configuration.")
 	fs.DurationVar(&s.InitialSyncTimeout, "initial-sync-timeout", s.InitialSyncTimeout,
 		"Timeout for initial resource sync.")
+
+	fs.StringVar(&s.ConfigDir, "config-dir", s.ConfigDir,
+		"directory to read config values from. Cannot be "+
+			"used in conjunction with federations or config-map flag.")
+	fs.DurationVar(&s.ConfigPeriod, "config-period", s.ConfigPeriod,
+		"period at which to check for updates in config-dir.")
 }
--- a/cmd/kube-dns/app/server.go
+++ b/cmd/kube-dns/app/server.go
@@ -53,15 +53,14 @@ func NewKubeDNSServerDefault(config *options.KubeDNSConfig) *KubeDNSServer {
 	}
 
 	var configSync dnsconfig.Sync
-	if config.ConfigMap == "" {
+	switch {
+	case config.ConfigMap != "":
+		glog.V(0).Infof("Using configuration read from ConfigMap: %v:%v", config.ConfigMapNs, config.ConfigMap)
+		configSync = dnsconfig.NewConfigMapSync(kubeClient, config.ConfigMapNs, config.ConfigMap)
+
+	default:
 		glog.V(0).Infof("ConfigMap not configured, using values from command line flags")
-		configSync = dnsconfig.NewNopSync(
-			&dnsconfig.Config{Federations: config.Federations})
-	} else {
-		glog.V(0).Infof("Using configuration read from ConfigMap: %v:%v",
-			config.ConfigMapNs, config.ConfigMap)
-		configSync = dnsconfig.NewSync(
-			kubeClient, config.ConfigMapNs, config.ConfigMap)
+		configSync = dnsconfig.NewNopSync(&dnsconfig.Config{Federations: config.Federations})
 	}
 
 	return &KubeDNSServer{
--- a/cmd/kube-dns/app/server.go
+++ b/cmd/kube-dns/app/server.go
@@ -54,12 +54,19 @@ func NewKubeDNSServerDefault(config *options.KubeDNSConfig) *KubeDNSServer {
 
 	var configSync dnsconfig.Sync
 	switch {
+	case config.ConfigMap != "" && config.ConfigDir != "":
+		glog.Fatal("Cannot use both ConfigMap and ConfigDir")
+
 	case config.ConfigMap != "":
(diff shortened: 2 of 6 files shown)


liggitt:
cc @thockin @MrHohn

cmd/kube-dns/app/server.go
```diff
@@ -53,15 +53,18 @@ func NewKubeDNSServerDefault(config *options.KubeDNSConfig) *KubeDNSServer {
...
+	case config.ConfigMap != "":
+		glog.V(0).Infof("Using configuration read from ConfigMap: %v:%v", config.ConfigMapNs, config.ConfigMap)
+		syncSource := dnsconfig.NewAPISyncSource(kubeClient, config.ConfigMapNs, config.ConfigMap)
+		configSync = dnsconfig.NewSync(syncSource)
+	case config.ConfigDir != "":
+		glog.V(0).Infof("Using configuration read from directory: %v", config.ConfigDir, config.ConfigPeriod)
+		syncSource := dnsconfig.NewFileSyncSource(config.ConfigDir, config.ConfigPeriod)
+		configSync = dnsconfig.NewSync(syncSource)
+	default:
 		glog.V(0).Infof("ConfigMap not configured, using values from command line flags")
```
GheRivero: Needs to be updated to inform that ConfigDir is neither configured

liggitt: updated

liggitt:
added tests for the file-based config

cmd/kube-dns/app/server.go
```diff
@@ -53,15 +53,18 @@ func NewKubeDNSServerDefault(config *options.KubeDNSConfig) *KubeDNSServer {
...
-	if config.ConfigMap == "" {
-		glog.V(0).Infof("ConfigMap not configured, using values from command line flags")
-		configSync = dnsconfig.NewNopSync(
-			&dnsconfig.Config{Federations: config.Federations})
-	} else {
-		glog.V(0).Infof("Using configuration read from ConfigMap: %v:%v",
-			config.ConfigMapNs, config.ConfigMap)
-		configSync = dnsconfig.NewSync(
-			kubeClient, config.ConfigMapNs, config.ConfigMap)
+	switch {
```
bowei: add 

```
if config.ConfigMap != "" && config.ConfigDir != "" {
  glog.V(0).Warnf("Both ConfigMap and ConfigDir set...")
}
```

cmd/kube-dns/app/server.go
```diff
@@ -53,15 +53,18 @@ func NewKubeDNSServerDefault(config *options.KubeDNSConfig) *KubeDNSServer {
...
+	case config.ConfigMap != "":
```
bowei: Do we want to prefer ConfigDir in this case?

cmd/kube-dns/app/options/options.go
```diff
@@ -166,4 +172,9 @@ func (s *KubeDNSConfig) AddFlags(fs *pflag.FlagSet) {
 			"dynamically adjustable configuration.")
 	fs.DurationVar(&s.InitialSyncTimeout, "initial-sync-timeout", s.InitialSyncTimeout,
 		"Timeout for initial resource sync.")
+
+	fs.StringVar(&s.ConfigDir, "config-dir", s.ConfigDir,
+		"directory to read config values from.")
```
bowei: add something like ". Only one of config-map or config-dir should be set"

pkg/dns/config/sync.go
```diff
@@ -47,50 +36,28 @@ type Sync interface {
...
+	Version string
+	Data    map[string]string
+}
+
+type SyncSource interface {
+	Once() (SyncResult, error)
+	Periodic() <-chan SyncResult
+}
+
 // NewSync for ConfigMap from namespace `ns` and `name`.
```
bowei: this comment needs to be updated

pkg/dns/config/sync.go
```diff
@@ -47,50 +36,28 @@ type Sync interface {
 	Periodic() <-chan *Config
 }
 
+type SyncResult struct {
+	Version string
+	Data    map[string]string
+}
+
+type SyncSource interface {
```
bowei: Seems like the same as Sync interface above? Why have another layer here?

pkg/dns/config/sync_api.go
```diff
@@ -0,0 +1,120 @@
+/*
```
bowei: can we name this sync_configmap.go? sync_api would refer to "the sync API" in my mind

pkg/dns/config/sync.go
liggitt: source just gets the data (from the API or from the disk)

that's logically distinct from deduping on version, parsing and validating that data into a config

bowei: Can we make SyncSource and SyncResult package private in that case

liggitt: no, they get returned and used by the app package

cmd/kube-dns/app/server.go
liggitt: added check for both below

liggitt:
comments addressed

pkg/dns/config/sync.go
bowei: I think SyncResult for sure does not need to be exposed. I just tried it with your patch and there isn't an issue.

It would be preferable to just have just expose NewConfigMapSync and NewFileSync constructors for SYnc and hide the SyncSource and SyncResult from outside altogether.

liggitt:
> It would be preferable to just have just expose NewConfigMapSync and NewFileSync constructors for Sync and hide the SyncSource and SyncResult from outside altogether.

sounds good, updated.

pkg/dns/config/sync_dir_test.go
```diff
@@ -0,0 +1,163 @@
...
+	if err != nil {
+		t.Fatalf("expected no error, got: %v", err)
+	}
+	if result.Version != "" || len(result.Data) != 0 {
+		t.Fatalf("expected empty version and data reading dir containing subdirs and dotfiles, got %#v", result)
+	}
+
+	// should return error if non-utf8 data is encountered
+	// https://en.wikipedia.org/wiki/UTF-8#Codepage_layout
+	if err := ioutil.WriteFile(filepath.Join(testDir, "binary"), []byte{192}, os.FileMode(0755)); err != nil {
```
MrHohn: Curious, does `data` in ConfigMap only contain UTF-8? I used to think that is only true for `key`, which got reflected as path here.

pkg/dns/config/sync_dir_test.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+	if err := ioutil.WriteFile(filepath.Join(testDir, "file2"), []byte("data2"), os.FileMode(0755)); err != nil {
+		t.Fatal(err)
+	}
+	result, err = source.Once()
+	if err != nil {
+		t.Fatalf("expected no error, got: %v", err)
+	}
+	expectedResult := syncResult{
+		Version: "d56592d659b731bdf7979b45644cba0ce80cc672ec7d30899757b4ad892c7360",
```
MrHohn: Is it possible to get this hash programmably?

pkg/dns/config/sync_dir_test.go
liggitt: it's `map[string]string`, not `map[string][]byte`, so it should only be utf-8 data

pkg/dns/config/sync_dir_test.go
liggitt: updated

MrHohn:
Thanks for the helps! LGTM @bowei

liggitt:
thanks! what's the timeframe or process for getting a version of the image made with this?

bowei:
uploading right now...

bowei:
I have plans to make the releases automated, but haven't gotten around to it. There are some issues with automation google_containers auth tokens which I have to resolve.

bowei:
gcr.io/google_containers/k8s-dns-dnsmasq-amd64:1.12.0
gcr.io/google_containers/k8s-dns-dnsmasq-arm:1.12.0
gcr.io/google_containers/k8s-dns-dnsmasq-arm64:1.12.0
gcr.io/google_containers/k8s-dns-dnsmasq-ppc64le:1.12.0
gcr.io/google_containers/k8s-dns-kube-dns-amd64:1.12.0
gcr.io/google_containers/k8s-dns-kube-dns-arm:1.12.0
gcr.io/google_containers/k8s-dns-kube-dns-arm64:1.12.0
gcr.io/google_containers/k8s-dns-kube-dns-ppc64le:1.12.0
gcr.io/google_containers/k8s-dns-sidecar-amd64:1.12.0
gcr.io/google_containers/k8s-dns-sidecar-arm:1.12.0
gcr.io/google_containers/k8s-dns-sidecar-arm64:1.12.0
gcr.io/google_containers/k8s-dns-sidecar-e2e-amd64:1.12.0
gcr.io/google_containers/k8s-dns-sidecar-e2e-arm:1.12.0
gcr.io/google_containers/k8s-dns-sidecar-e2e-arm64:1.12.0
gcr.io/google_containers/k8s-dns-sidecar-e2e-ppc64le:1.12.0
gcr.io/google_containers/k8s-dns-sidecar-ppc64le:1.12.0

PR #40, opened 2017-01-24
Lock client-go at release 2.0

### LINKED ISSUE #35: kube-dns should use versioned client-go
kube-dns should use a versioned copy of the client-go repository, rather than HEAD

cmluciano:
WIP lock client-go to release v2.0.0-alpha.1


bowei:
Could you split the single commit into two? One commit for non-vendor directory changes, another with only vendor directory changes.

something like:
```sh
git rebase -i HEAD^ # edit the commit
git reset HEAD^
git add vendor/
git commit -m "Update vendor to client-go v1.5"
git rebase --continue
```

cmluciano:
For sure, I can already tell that I borked at least one dependency and client-go update to sync.

I will work on setting up a fresh dev env on Linux since macOS is not very conducive to some of these test runs.

cmluciano:
@bowei I'm still having a difficult time with this. It appears that Godeps removes 3 pkgs that are were originally vendored.

```
vendor/github.com/googleapis/gax-go/
vendor/bitbucket.org/ww/goautoneg/
vendor/golang.org/x/net/
vendor/google.golang.org/grpc/
```

I'm not sure why Godep removes them, but the travis build suggests that they are still needed.

My steps are:
1. Checkout 1.5 version of client-go
2. Make necessary code changes
3. Use script (have tried w/o script) to save dependencies after removing Godeps/ and vendor/
4. Assess changes
5. make build 

```
make build
building : bin/amd64/kube-dns
vendor/cloud.google.com/go/internal/retry.go:21:2: cannot find package "github.com/googleapis/gax-go" in any of:
	/go/src/k8s.io/dns/vendor/github.com/googleapis/gax-go (vendor tree)
	/usr/local/go/src/github.com/googleapis/gax-go (from $GOROOT)
	/go/src/github.com/googleapis/gax-go (from $GOPATH)
vendor/golang.org/x/text/cases/map.go:16:2: cannot find package "golang.org/x/text/internal" in any of:
	/go/src/k8s.io/dns/vendor/golang.org/x/text/internal (vendor tree)
	/usr/local/go/src/golang.org/x/text/internal (from $GOROOT)
	/go/src/golang.org/x/text/internal (from $GOPATH)
make: *** [bin/amd64/kube-dns] Error 1
```

I'm not sure if these are not needed in 1.5 and some import paths are not stripping them out.

bowei:
let me take a look

liggitt:
client-go release numbers are not related to kubernetes release numbers.

see https://github.com/kubernetes/client-go/tree/release-2.0#versioning and https://github.com/kubernetes/client-go/tree/release-2.0#compatibility-matrix

> 1.4 and 1.5 branch keep the top-level folders so they do not break the import lines of existing users. These top-level folders are deprecated and are removed from the master branch and future branches.

cmluciano:
@liggitt My revised steps have been 

```
rm -rf client-go
go get client-go and checkout the v1.5 release
update import paths to client-go and then run our godep save script
```

Does this seem correct?
Should we use apimachinery for [metav1 import paths](https://github.com/kubernetes/dns/blob/master/pkg/dns/dns.go#L28)? The metav1 paths seem very different in 1.5. 

I can open an issue on client-go if that is helpful.

liggitt:
I would avoid adopting 1.5 client-go patterns at this point... client and apimachinery separation work is in progress and will look much more like the long-term plan in 1.6

bowei:
discussed with @cmluciano  -- we will go with release-2.0 branch for the client, this avoids the legacy package structure

cmluciano:
The travis builds are complaining about the lack of[k8s.io/client-go/pkg/apis/meta/v1](https://travis-ci.org/kubernetes/dns/builds/196677328#L261).

Some people added apimachinery directly to solve this problem, but it was eventually added explicitly as an import in client-go.

The alpha release seems to use the unversioned path for a lot of these things, but since this change is about locking to a stable branch, I wanted to confirm which direction we should go.

bowei:
@cmluciano -- let me see what the apimachinery people recommend

bowei:
@caesarxuchao -- We have some questions regarding how to best use the client-go repo. We are thinking about tracking release-2.0 as a dependency, but there seems to be some missing bits (see above "k8s.io/client-go/pkg/apis/meta/v1"). What is the recommended approach to using the client-go repo?

caesarxuchao:
If the dns project doesn't require new API resources defined after Kubernetes release 1.5, then I recommend tracking client-go release-2.0. The code in client-go release-2.0 is tracking the client related code in kubernetes release 1.5.

"k8s.io/client-go/pkg/apis/meta/v1" was created after kubernetes release 1.5. Its contents were moved from "k8s.io/client-go/pkg/api" or "k8s.io/client-go/pkg/api/unversioned".

cmluciano:
@bowei 

Did some swapping around and it feels like we started to shift pretty far down the path of features that are not cherrypicked into the client-go release-2 branch. The main problems are in the sync_configmap.go section here https://github.com/kubernetes/kubernetes/blob/19be1d250426ddf9842d9d66c69f91df05b8b28f/pkg/dns/config/sync.go#L49-L66. ListOptions issues are pretty commonplace with the shift to client.go and the kubernetes.Interface. 

I will push up a build with some swaps of unversioned with the metav1 alias so that the travis build will detail what I mean. 

I'm hoping that I am just missing something obvious.

bowei:
@cmluciano, after taking a look at the diff, I'm inclined to wait for the 1.6 stable release to be tagged and not have to do a bunch of changes that will most likely need to be reverted when we do the version bump in the future. We are almost at the midpoint of the 1.6 cycle and there are no (I hope) burning issues with the client-go versioning at the moment.

What do you think?

bowei:
@caesarxuchao -- when can we expect the release tag to be updated? I know client-go no longer tracks the Kubernetes release numbering...

cmluciano:
@bowei I fully agree.

caesarxuchao:
I'm waiting for https://github.com/kubernetes/kubernetes/pull/40785 to be merged, then I'll cut the client-go 2.0 stable version.

PR #44, opened 2017-01-31
Change shebang to sh instead of bash

cmluciano:
Bash is not available in the golang container that this script runs in.

### DIFF
--- a/build/dep.sh
+++ b/build/dep.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 # Copyright 2017 The Kubernetes Authors.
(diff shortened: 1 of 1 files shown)


cmluciano:
I have not had any unintended sideffects on Ubuntu 16.04. I also verified with the [checkbashism tool](http://manpages.ubuntu.com/manpages/zesty/man1/checkbashisms.1.html).

```
checkbashisms /vagrant/build/dep.sh 
could not find any possible bashisms in bash script /vagrant/build/dep.sh
```

bowei:
thanks -- merged

PR #53, opened 2017-02-10
Update README.md

ronaldpetty:
Changed verbiage, we build images not containers.  Didn't change the make command though as I don't fully appreciate what might go wrong if that changes (assuming automatic processes do `make containers`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,8 +20,8 @@ This is the repository for [Kubernetes DNS](http://kubernetes.io/docs/admin/dns/
 | ---- | ---- |
 |all, build | build all binaries |
 |test       | run unit tests |
-|containers | build the containers |
-|push       | push containers to the registry |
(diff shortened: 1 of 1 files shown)


cmluciano:
LGTM

ronaldpetty:
I agree @bowei for all or nothing, just was avoiding the risk.  I understand the confusing bit, but I personally rather have more correct docs that our confusing than incorrect docs that work.  Adhore magic.  I don't know the impact, but if you think its has an upside that trumps (yep) the people who already use it and don't want to get blasted by a code update then I am happy to give it a go!

bowei:
@ronaldpetty -- I think the change should be a simple `s/container/image/`, just have to watch out for capitalization (container, CONTAINER). The change should be limited to Makefile, rules.mk, images/*/Makefile.

PR #62, opened 2017-02-19
Use default config if none exist at startup

bowei:
Also adds some glog flush() to ensure we see more of the log output


cmluciano:
👍🏽

PR #63, opened 2017-02-23
Use protobufs for communication with apiserver

wojtek-t:
@bowei @thockin


bowei:
looks like CI failed?

wojtek-t:
@bowei - sorry, yes it's gofmt.

But I wasn't able to update the PR (I created it from UI and something didn't work for me). So to not waste too much time on it, closing in favor of https://github.com/kubernetes/dns/pull/64

PTAL  on that one.

PR #78, opened 2017-04-19
Fix confusing logging when initialize server

MrHohn:
The logging we have when initializing services and endpoints is confusing:
```
DNS server not ready, retry in 500 milliseconds
```

Modify the message a bit to clarify what it is doing.


bowei:
merge when ready

MrHohn:
Thanks, all test passed, merging.

PR #79, opened 2017-04-20
Fix race condition in reverseRecordMap access

MrHohn:
Fix kubernetes/kubernetes#44711.

We don't have a explicit lock for `reverseRecordMap`, all the R/Ws are protected through `cacheLock`, so I followed the same pattern.

@bowei 
cc @etiennetremel


bowei:
Good catch, I thought I fixed all of these. Can you do a brief scan to see that the all reads hold the lock?

MrHohn:
Did a quick check, two possible concurrent reads in https://github.com/kubernetes/dns/blob/master/pkg/dns/dns.go#L714 and https://github.com/kubernetes/dns/blob/master/pkg/dns/dns.go#L724.

MrHohn:
Updated PR, PTAL

weiwei04:
Encounter this problem, looking forward to see this pr be merged.

PR #87, opened 2017-05-10
Use BSD-compatible sed command in dnsmasq Makefile.

### LINKED ISSUE #86: dnsmasq container fails to build on OS X (incompatible with BSD sed)
The Makefile for the dnsmasq container contains invocations in the style of:

```bash
sed -i "s/pattern/replacement/" Dockerfile
```
This is incompatible with BSD `sed`, whose `-i` flag takes a required argument. It generates the following:
```
sed: 1: "Dockerfile": extra characters at the end of D command
```
Note that it's trying to parse the string `Dockerfile` as a `sed` command.

I'll have a patch shortly.

jkinkead:
Fixes #86 .

This invocation should work in both BSD & Linux `sed`. Note that I remove the backup file being created.


thockin:
@bowei is this repo using bots or manual merge?

bowei:
manual merge

PR #101, opened 2017-05-30
Fix PTR Record generation

sadlil:
Ref : https://github.com/kubernetes/dns/pull/25#issuecomment-304734753

### DIFF
--- a/pkg/dns/dns.go
+++ b/pkg/dns/dns.go
@@ -261,7 +261,7 @@ func assertIsService(obj interface{}) (*v1.Service, bool) {
 
 func (kd *KubeDNS) newService(obj interface{}) {
 	if service, ok := assertIsService(obj); ok {
-		glog.V(2).Infof("New service: %v", service.Name)
+		glog.V(3).Infof("New service: %v", service.Name)
 		glog.V(4).Infof("Service details: %v", service)
 
 		// ExternalName services are a special kind that return CNAME records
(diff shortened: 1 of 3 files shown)


pkg/dns/dns.go
```diff
@@ -523,6 +523,23 @@ func getHostname(address *v1.EndpointAddress) (string, bool) {
 	return "", false
 }
 
+func (kd *KubeDNS) ensurePodSubdomain(address *v1.EndpointAddress, svc *v1.Service) bool {
```
thockin: How frequently does this run?  I mean, it runs on every pod that needs it, but what is the period for refresh?

I'm worried about swamping the apiserver when there are lots of DNS replicas active, but we really can't watch pods.

@bowei @smarterclayton

pkg/dns/dns_test.go
```diff
@@ -427,6 +429,7 @@ func TestNamedHeadlessServiceEndpointUpdate(t *testing.T) {
 
 func TestNamedHeadlessServiceEndpointDelete(t *testing.T) {
```
thockin: Add a testcase for a en endpoint that requests reverse DNS but the pod doesn't confirm?

pkg/dns/dns.go
feiskyer: Does adding the subdomain to endpoints make sense for this?  (not sure whether it is general enough for endpoints having a subdomain.)

thockin:
@bowei shoot - this fell through again.

chrislovecnm:
@thockin / @sadlil what is the status on this?  How do we not wait for 1.9 for this???

bowei:
@chrislovecnm Are you saying to put in 1.8 or fix in 1.9?

chrislovecnm:
@bowei we need a fixed container ASAP, but when can we get this released?

PR #105, opened 2017-06-12
Bump to Go version 1.8

cmluciano:
Signed-off-by: Christopher M. Luciano <cmluciano@us.ibm.com>

PR #108, opened 2017-06-13
Update dnsmasq image to use go1.7.6 and alpine:3.6

ixdy:
Updated golang dependency includes the fix for https://github.com/golang/go/issues/20040.
Updated alpine dependency includes fixes for CVE-2016-9841 and CVE-2016-9843.

Note that we should probably build a new dnsmasq image before building the rest, since dnsmasq-nanny depends on it.

@MrHohn @bowei 

x-ref https://github.com/kubernetes/kubernetes/issues/47386

### DIFF
--- a/images/dnsmasq/Makefile
+++ b/images/dnsmasq/Makefile
@@ -13,22 +13,22 @@
 # limitations under the License.
 
 VERSION ?= $(shell git describe --tags --always --dirty)
-REGISTRY ?= gcr.io/google_containers
+REGISTRY ?= gcr.io/google-containers
 ARCH ?= amd64
 DNSMASQ_VERSION ?= dnsmasq-2.76
 CONTAINER_PREFIX ?= k8s-dns
 
 ALL_ARCH := amd64 arm arm64 ppc64le s390x
 IMAGE := $(CONTAINER_PREFIX)-dnsmasq-$(ARCH)
-COMPILE_IMAGE := gcr.io/google_containers/kube-cross:v1.7.4-1
+COMPILE_IMAGE := gcr.io/google-containers/kube-cross:v1.7.6-k8s1.6-0
 OUTPUT_DIR := _output/$(ARCH)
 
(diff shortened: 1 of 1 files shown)


images/dnsmasq/Makefile
```diff
@@ -13,19 +13,19 @@
 # limitations under the License.
 
 VERSION ?= $(shell git describe --tags --always --dirty)
-REGISTRY ?= gcr.io/google_containers
+REGISTRY ?= gcr.io/google-containers
```
MrHohn: So we are consolidating images to `google-containers` in future? Just notice that cluster-proportional-autoscaler is still using `google_containers`.

ixdy: yeah, they're basically equivalent as far as docker is concerned, but various utilities prefer the real project name (`google-containers`) instead of the docker hack (`google_containers`), so I've been slowly updating references as I come across them.

one of these days I might just do a large fixit.

ixdy:
hm, looking into the arm build failure.

ixdy:
OK, fixed the build by changing arm from arm-linux-gnueabi to arm-linux-gnueabihf. I hope this is OK? @luxas

MrHohn:
Will go ahead and merge it tonight if no more comments :)

cmluciano:
I had a similar PR that bumped to go 1.8 to match kubernetes/kubernetes here https://github.com/kubernetes/dns/pull/105. Should I close this in favor of sticking with 1.7.X ?

MrHohn:
@cmluciano I think this PR is more about fixing CVEs in kube-dns images and that needs to get into 1.7 release. Let's keep #105 as we are not sticking with 1.7.x forever.

bowei:
Do we need to push a new tag and release and update the reference in the main repo?

ixdy:
@bowei Yeah, we should update the dnsmasq image and then ideally the rest of the k8s-dns images, too.

PR #120, opened 2017-06-29
Add some error and info log for debug

zhxcai:


zhxcai:
@cmluciano Don't have any errors these. 

Just because when test the mysql galera example(https://github.com/kubernetes/kubernetes/tree/master/examples/storage/mysql-galera) , I get the pod name (ex: pxc-node1 ) dns record not is the pod ip, So we want to debug dns watched result.

Finally, I found out the dns record IP is the service cluster ip, and the name created of service is same to pod name.

pkg/dns/dns.go
```diff
@@ -390,7 +397,12 @@ func (kd *KubeDNS) handleEndpointDelete(obj interface{}) {
 	}
 
 	svc, err := kd.getServiceFromEndpoints(endpoints)
-	if svc != nil && err == nil {
+	if err != nil {
+		glog.Errorf("Error get service from endpoints %s: %v",
```
bowei: Errorf("Error from getServiceFromEndpoints(%v): %v", endpoints.Name, err)

pkg/dns/dns.go
```diff
@@ -510,6 +522,8 @@ func (kd *KubeDNS) generateRecordsForHeadlessService(e *v1.Endpoints, svc *v1.Se
 	kd.cacheLock.Lock()
 	defer kd.cacheLock.Unlock()
 	for endpointIP, reverseRecord := range generatedRecords {
+		glog.V(4).Infof("generateRecordsForHeadlessService: save endpointIP %q with reverseRecord value: %s",
```
bowei: "Adding endpointIP %q to reverseRecord %+v"

zhxcai:
@bowei Thanks for your advice.

Issue #116, opened 2017-06-26
Only one A record set for headless service with pods having single hostname.


spil-dennis:
/kind bug

** What happened **
When a headless service is created to point to pods which share a single hostname, (which happens, for example, when the hostname field was set in the template of a deployment/replicaset)
- Only one A record is returned for the service DNS name
- A pod DNS name is generated based on this host name, which points to a single pod

** What was expected to happen **
- Return A records for all available endpoints on the service DNS name
- Not sure what the correct behaviour should be for the pod dns name, either also return multiple A records, or don't create at all.

Seems this has to do with the following code:
https://github.com/kubernetes/dns/blob/master/pkg/dns/dns.go#L490

Then the endpointName will be equal for any pod in the service which has the same hostname, so the entry in subCache will be overwritten.

** How to reproduce **

Apply the following spec:

```
apiVersion: v1
kind: List
items:
- apiVersion: extensions/v1beta1
  kind: Deployment
  metadata:
    name: depl-1
  spec:
    replicas: 2
    template:
      metadata:
        labels:
          app: depl-1
      spec:
        hostname: depl-1-host
        subdomain: depl-1-service
        containers:
        - name: test
          args:
          - bash
          stdin: true
          tty: true
          image: debian:jessie
- apiVersion: v1
  kind: Service
  metadata:
    name: depl-1-service
  spec:
    clusterIP: None
    selector:
      app: depl-1
    ports:
    - port: 5000

```

Resolving the hostnames gives back but a single A record.
```
# host depl-1-host.depl-1-service.default.svc.cluster.local
depl-1-host.depl-1-service.default.svc.cluster.local has address 10.56.0.140
# host depl-1-service.default.svc.cluster.local
depl-1-service.default.svc.cluster.local has address 10.56.0.140
```

PTR records ARE being created for all the pods, all resolving back to the single hostname. This is expected behaviour.

thockin:
This is a real bug that I expect people to hit. :(
/lifecycle frozen

StephanX:
Yep, just ran into this today, while trying to use a headless service to permit direct access to pods in a replicaset.  I tried working around it by omitting the hostname from pod definition, but kubedns just ignored the pod all together.

[...]

PR #123, opened 2017-07-07
Handle same hostname for multiple pods in a headless service

guillaumerose:
This code improves the handling of headless services pointing at pods sharing
the same hostname.

 + For each hostname, A records should point at the list of pods IPs.
   Currently, only the last pod gets its A record.
 + For the service, a A record should be created for each pod.

Fixes https://github.com/kubernetes/dns/issues/116

cc @dgageot

### DIFF
--- a/pkg/dns/dns.go
+++ b/pkg/dns/dns.go
@@ -489,7 +489,17 @@ func (kd *KubeDNS) generateRecordsForHeadlessService(e *v1.Endpoints, svc *v1.Se
 			if hostLabel, exists := getHostname(address); exists {
 				endpointName = hostLabel
 			}
-			subCache.SetEntry(endpointName, recordValue, kd.fqdn(svc, endpointName))
+			if addresses := addressesForHostname(e, endpointName); len(addresses) > 1 {
+				endpointCache := treecache.NewTreeCache()
+				for _, address := range addresses {
+					recordValue, endpointName := util.GetSkyMsg(address.IP, 0)
+					endpointCache.SetEntry(recordValue.Host, recordValue, kd.fqdn(svc, endpointName, endpointName))
+					subCache.SetEntry(recordValue.Host, recordValue, kd.fqdn(svc, endpointName))
+				}
+				subCache.SetSubCache(endpointName, endpointCache)
+			} else {
+				subCache.SetEntry(endpointName, recordValue, kd.fqdn(svc, endpointName))
+			}
 			for portIdx := range e.Subsets[idx].Ports {
 				endpointPort := &e.Subsets[idx].Ports[portIdx]
 				if endpointPort.Name != "" && endpointPort.Protocol != "" {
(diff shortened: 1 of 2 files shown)


thockin:
I know the spec says it should return multiple A records in this case, but I just want to think it through.  Is this the correct behavior?  It's less surprising than dropping all-but-one, though I could argue that using the same hostname twice is a bug on the user's end.

So is the spec wrong or is the code wrong?  If we implement this - what else might break?  Are there apps that actually expect to use DNS as "phantom elimination"?  I feel the current behavior is under-specified to rely on that, but the implementation is reasonably deterministic.  

@bowei @johnbelamaric @smarterclayton @erictune @kubernetes/sig-apps-misc

erictune:
@kow3ns

erictune:
Using the same hostname is not a bug.  

Spreading may be soft rather than hard.  

Someone may want to bring up a size-3 etcd in their 2-node cluster as part of a test.  They don't need HA, they just want to reuse the config.  That should work as similarly as possible to the "normal" case where the pods are properly spread.

thockin:
@erictune I am not clear on your reply.  Let me provide an example and you tell me which behavior is considered correct.

T0: User somehow creates a set of pods that have hostnames "app-1", "app-2", "app-3".  DNS lookups of "service" return 3 distinct A records.  DNS lookups of "app-[123].service" each return one distinct A record.

T1: Some management event happens and, through cut-paste or similar, the app-3 pod is replaced with a new pod that has hostname "app-2".  DNS lookups of "service" return 3 distinct A records.  Question: should a DNS lookup of "app-2" return a single A record (one of the 2 pods, chosen somehow) or should it return 2 A records?

Today it returns 1 record.

thockin:
Another note: if we decide the spec is wrong, and that we should only serve one A record, we have a bug.  When you hit this overlap, the IP of the last pod created "wins".  If that pod goes away, we do not update the PTR to the other pod with the same hostname.

erictune:
Please disregard previous comment.  I was not aware of how the hostname
override behavior affected DNS.

I don't know of any use cases that break either way.  Ken may know some.

dgageot:
Thanks @thockin for the feedback.

When we prepared this PR, we tried real hard to not break existing things and we considered this whole "phantom elimination" couldn't be considered a feature given what's in the spec. But I understand that nonetheless, it *might* be a breaking change.

What do you think we can do to mitigate that risk?

johnbelamaric:
@chrisohaver I believe in CoreDNS we include all IPs. This is consistent with ordinary DNS behavior - it is legal in DNS to have multiple A records with the same name and different IPs.

guillaumerose:
We fixed the CLA check. Any chance to get this merged ? Should we do something ?
Thanks !

dgageot:
ping @thockin

guillaumerose:
Any update on this ?

PR #133, opened 2017-08-07
[POC] Support specifying namespace

### LINKED ISSUE #132: Add support for serving on specified namespaces
kube-dns is serving on all namespaces today. As we are moving on to support multiple networks, pods on different networks may not access each other, e.g. in a multi-tenant environment. 

So I propose to support serving on specified namespace in kube-dns, so as to conform such use cases.

cc/ @bowei

feiskyer:
Closes #132.


feiskyer:
cc/ @bowei

feiskyer:
Close the issue since it's not our recommended way.

PR #135, opened 2017-08-11
code refactor

andyxning:
This PR refactor the code to use `metricsClient` struct field instead of a local variable `client`.

/cc @bowei


bowei:
probably should be the other way, it looks like the struct variable is never used anywhere else and should be removed.

andyxning:
Yep. Actually i refactor it by removing the useless struct firstly. But after a thought, i think `probes`, `metricsClient` and `options` struct field is not necessary indeed. So i we delete `metricsClient` instead of `use it, may be `options` and `probes` should also be deleted.

andyxning:
Friendly ping @bowei

PR #136, opened 2017-08-12
Fix: update `kube2sky` to `kube-dns` in comment of server.go 

fisherxu:
I think the  `kube2sky` should be  `kube-dns`, so I fix it.


fisherxu:
ping @bowei help review this update, thanks!

PR #146, opened 2017-08-24
fix make file

andyxning:
remove useless build volume mapping when build binaries.

By removing `-v $$(pwd)/bin/$(ARCH):/go/bin`, we can avoid an empty dir  `bin/ARCH/linux_ARCH`.

/cc @bowei

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,14 +19,14 @@
 # Golang package.
 PKG := k8s.io/dns
 
-# List of binaries to build. You must have a matching Dockerfile.BINARY
-# for each BINARY.
+# List of binaries to build.
 BINARIES := \
     e2e \
     ginkgo \
     sidecar-e2e
 
 # List of binaries to build that are containerized and pushed.
+# You must have a matching Dockerfile.BINARY for each BINARY.
 CONTAINER_BINARIES := \
     dnsmasq-nanny \
     kube-dns \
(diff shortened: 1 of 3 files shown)


thockin:
Needed for cross-builds, if I recall.

andyxning:
@thockin After a deep look, you're right. It is mainly used to get none amd64 binaries. 

In order to make the volume mount more clear and maintainable, I suggest we always only mount `/go/bin/linux_ARCH` and use `GOBIN` to effectively workaround the problem to get `amd64` binaries.

andyxning:
@bowei @thockin PTAL.

bowei:
Can you explain in detail what this does change? What is the problem you are trying to solve?

andyxning:
@bowei

As stated in the PR description. For now, there will be an empty `bin/ARCH/linux_ARCH` dir after the `make build`. It is useless, imo. By removing `-v $$(pwd)/bin/$(ARCH):/go/bin`, we can avoid an empty dir `bin/ARCH/linux_ARCH`.

bowei:
That directory is created in bin/ which removed as part of `make clean`. Is this causing a problem with the build?

andyxning:
No. There is no problem in fact. :) 

Just to make the `bin` directory more cleaner.

andyxning:
ping @bowei

andyxning:
cc @bowei

bowei:
ok, lgtm

andyxning:
@bowei Should we merge this PR? :)

PR #148, opened 2017-09-01
Add support for nameservers with IPv6 addresses

### LINKED ISSUE #147: Add support for nameservers with IPv6 addresses
If kube-dns is run on a minion that has a nameserver listed in
/etc/resolv.conf with an IPv6 address, then the dnsmasq nanny container
dies shortly after reading this resolv.conf file, and the liveness probe
never indicates a healthy pod (pod is restarted after a few minutes).

leblancd:
This change fixes a couple of places in the kube-dns code where
IPv6 addresses are not being handled correctly for nameservers found
in a minion's resolv.conf file.

This change is needed in order to get kube-dns to operate in a
Kubernetes cluster that is IPv6-only. See Kubernetes issue 51371
(https://github.com/kubernetes/kubernetes/issues/51371).

fixes #147

### DIFF
--- a/cmd/kube-dns/app/server.go
+++ b/cmd/kube-dns/app/server.go
@@ -22,7 +22,6 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
-	"regexp"
 	"strconv"
 	"strings"
 	"syscall"
@@ -172,6 +171,25 @@ func validateHostAndPort(hostAndPort string) error {
 	return nil
 }
 
+// validateNameServer validates a configured name server and
+// appends a ":53" if a :port is not included. IPv6 addresses are also
+// enclosed in square brackets if not already so enclosed.
+func validateNameServer(nameServer string) (string, error) {
+	if ip := net.ParseIP(nameServer); ip != nil {
+		if ip.To4() != nil {
+			// IPv4 without port
+			nameServer = nameServer + ":53"
+		} else {
+			// IPv6 without port
+			nameServer = "[" + nameServer + "]:53"
+		}
+		return nameServer, nil
+	}
+	// Assume it's IP:port
+	err := validateHostAndPort(nameServer)
+	return nameServer, err
+}
+
 func (d *KubeDNSServer) startSkyDNSServer() {
 	glog.V(0).Infof("Starting SkyDNS server (%v:%v)", d.dnsBindAddress, d.dnsPort)
 	skydnsConfig := &server.Config{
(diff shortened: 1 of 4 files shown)


danehans:
@leblancd can you rebase? @bowei we need this patch for IPv6 e2e tests that are being finalized. Are you able to review this PR? If not do you recommend someone else?

cmd/kube-dns/app/server.go
```diff
@@ -172,6 +171,28 @@ func validateHostAndPort(hostAndPort string) error {
 	return nil
 }
 
+// validateNameServer validates a configured name server and
+// appends a ":53" if a :port is not included. IPv6 addresses are also
+// enclosed in square brackets if not already so enclosed.
+func validateNameServer(nameServer string) (string, error) {
+	if ip := net.ParseIP(nameServer); ip != nil {
```
bowei:
```
ip := net.ParseIP(...)
if ip == nil {
  // probably should return an error here
}

if ip.To4() != nil {
} else {
}

if nameServer[0] == '[' && nameServer[len(nameServer) -1] == ']' {
  // you need to validate the string inside is valid ipv6?
}
```

leblancd: @bowei : We can't return an error when net.ParseIP(...) returns nil, since we need to fall through to 'err := validateHostAndPort(nameServer)'. The logic here is that if it's not a valid (stand-alone, without a port) IP addr, then we need to assume it's a host + port string, and validate it with validateHostAndPort() (and that function will vet the string and return an error if it's abnormal).
Re. "you need to validate the string inside is valid ipv6?"... that is a very good point, but looking at this now, this check for an IPv6 address inside brackets should not be here. Brackets should not be seen around "stand-alone" IPv6 address... we should only see brackets around IPv6 addresses when it appears as part of an IP:port format, or it appears as part of a URL. I'll delete this check for IPv6 inside brackets.

danehans:
@bowei does the updated PR from @leblancd address your questions?

bowei:
Sorry about the delay -- I'll take a look today

danehans:
@bowei are there any other changes required for this PR to get approved? This is one of the IPv6 PRs needed for the 1.9 release.

bowei:
lgtm thanks for your patience

danehans:
@bowei thank for your help. Can you confirm that we need to follow the [Release Process](https://github.com/kubernetes/dns#release-process) and submit a PR to update the kube-dns image?

bowei:
I will do the update to 1.14.7. Unfortunately, kube-dns is still a manual process.

danehans:
Much appreciated. Can you ping me after updating to 1.14.7? I need to submit a follow-on PR to have kubeadm support 1.14.7.

bowei:
https://github.com/kubernetes/kubernetes/pull/54443

PR #163, opened 2017-11-07
set user-agent

weiwei04:
set user-agent for client-go

before

```
client="kube-dns/v0.0.0 (linux/amd64) kubernetes/$Format"
```

after

```
client="kube-dns/1.14.4-2-g5584e04"
```

Signed-off-by: Wei Wei <weiwei.inf@gmail.com>

### DIFF
--- a/cmd/kube-dns/app/server.go
+++ b/cmd/kube-dns/app/server.go
@@ -22,6 +22,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"runtime"
 	"strconv"
 	"strings"
 	"syscall"
@@ -38,6 +39,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/dns/pkg/version"
 )
 
 type KubeDNSServer struct {
@@ -103,8 +105,13 @@ func newKubeClient(dnsConfig *options.KubeDNSConfig) (kubernetes.Interface, erro
 	}
 	// Use protobufs for communication with apiserver.
 	config.ContentType = "application/vnd.kubernetes.protobuf"
+	config.UserAgent = userAgent()
 
 	return kubernetes.NewForConfig(config)
+}
+
+func userAgent() string {
+	return fmt.Sprintf("kube-dns/%s (%s/%s)", version.VERSION, runtime.GOOS, runtime.GOARCH)
 }
 
 func (server *KubeDNSServer) Run() {


cmd/kube-dns/app/server.go
```diff
@@ -103,6 +104,7 @@ func newKubeClient(dnsConfig *options.KubeDNSConfig) (kubernetes.Interface, erro
 	}
 	// Use protobufs for communication with apiserver.
 	config.ContentType = "application/vnd.kubernetes.protobuf"
+	config.UserAgent = "kube-dns/" + version.VERSION
 
 	return kubernetes.NewForConfig(config)
```
MrHohn: Instead of generating a customized one, what about using the library?
```go
return kubernetes.NewForConfig(rest.AddUserAgent(config, "kube-dns"))
```

Ref: https://github.com/kubernetes/client-go/blob/18d0325d5c7c66f0b14004a78260ecfd4cb6f7e6/rest/config.go#L396-L400

weiwei04: the AddUserAgent logic is
```golang
	fullUserAgent := DefaultKubernetesUserAgent() + "/" + userAgent
	config.UserAgent = fullUserAgent
	return config
```

and `DefaultKubernetesUserAgent()` already include the process name which is `kube-dns`
```golang
// DefaultKubernetesUserAgent returns a User-Agent string built from static global vars.
func DefaultKubernetesUserAgent() string {
	return buildUserAgent(
		adjustCommand(os.Args[0]),
		adjustVersion(version.Get().GitVersion),
		gruntime.GOOS,
		gruntime.GOARCH,
		adjustCommit(version.Get().GitCommit))
}
```

for kubernetes repo component like kube-apiserver, kube-scheduler etc, they compiled from kubernetes repo, and the build script will fill in client-go `gitVersion` `gitCommit`
see k8s.io/kubernetes/hack/lib/version.sh
```bash
    ldflags+=($(kube::version::ldflag "gitVersion" "${KUBE_GIT_VERSION}"))
```

which the user-agent would like 

`kube-apiserver/v1.8.2 (linux/amd64) kubernetes/bdaeafa`

shall we use the same logic to fill in client-go `gitVersion`(which is version.Get().GitVersion) `gitCommit`(which is version.Get().GitCommit) to create a user-agent like 

`kube-dns/1.14.4-2 (linux/amd64) kubernetes/g5584e04` 

or just 

`kube-dns/1.14.4-2-g5584e04`.

I'd prefer `kube-dns/1.14.4.2-g5584e04`, since it is shorter and contains all the info we need.

what do you think @MrHohn :)

MrHohn: Thanks for your investigation :) I'd lean more towards keeping kube-dns consistent with the other k8s components, also `GOOS` and `GOARCH` feel like something that might be helpful to include.

To do so, would it be as simple as a function call to `DefaultKubernetesUserAgent()`? Let me know if it is more complicated.

weiwei04: To do so, we need to use build script set `/vendor/k8s.io/client-go/pkg/version.gitCommit` and `/vendor/k8s.io/client-go/pkg/version.gitVersion` with ldflags as `DefaultKubernetesUserAgent()` depends on client-go/pkg/version, I am not sure if it is the only way. But I think we should relay on the exposed config.UserAgent(since it's kind of a api) instead of unexposed golang var with ldflags(it may change without further notice).

GOOS and GOARCH is also helpful, how about I write a helper function in dns repo to create a user-agent in a form like

`kube-dns/1.14.4-2-g5584e04 (linux/amd64)`?

MrHohn: Actually, took a another look, using those `UserAgent` funcs in client-go seems a bit heavy now. What you suggested sounds good. Could you make the format as `kube-dns/1.14.4-2-g5584e04 (linux/amd64)`? Thanks :)

weiwei04:
changed to `kube-dns/1.14.7-3-ga511f49 (linux/amd64)` ptal @MrHohn :)

PR #167, opened 2017-11-20
enable skyDNS round robin A/AAAA records

wu8685:
After issue #123 fixed, I hope we can enable skyDNS round robin to support load balance for headless service.


wu8685:
Shall we consider this PR to enable round robin in order to support LB for multiple pods behind a headless service ?

thockin:
What is the end result of this?  If you have a headless service with 3 endpoints, every query should get all 3 addresses, no?  Does this just shuffle them?

miekg:
[ Quoting <notifications@github.com> in "Re: [kubernetes/dns] enable skyDNS ..." ]
>What is the end result of this?  If you have a headless service with 3 endpoints, every query should get all 3 addresses, no?  Does this just shuffle them?

I believe so, yes.

wu8685:
Yes, it dose shuffle:)

However, it is not enough.
When we use the dns only has shuffle to LB the headless svc, the backend server does not be hit equally.

I have done some test to count the backend server hits. The result is listed below:
```
// enable RR
ip: 10.212.151.227, count: 530
ip: 10.212.151.218, count: 470

// disable RR
ip: 10.212.151.218, count: 873
ip: 10.212.151.227, count: 127
```

miekg:
[ Quoting <notifications@github.com> in "Re: [kubernetes/dns] enable skyDNS ..." ]
>Yes, it dose shuffle:)
>
>However, it is not enough.
>When we use the dns only has shuffle to LB the headless svc, the backend server does not be hit equally.
>
>I have done some test to count the backend server hits. The result is listed below:
>```
>// enable RR
>ip: 10.212.151.227, count: 530
>ip: 10.212.151.218, count: 470
>
>// disable RR
>ip: 10.212.151.218, count: 873
>ip: 10.212.151.227, count: 127
>```

this just randomizes, has no concept of load or any other external factors.

wu8685:
Is it a good ideal to export the round-robin as a kube-dns configuration ?
Users can decide themselves.

thockin:
Shuffling seems OK.  I assumed that clients would randomly pick from a
response set, but maybe they tend to take the first item.  Low cost to
shuffle, I guess.

The real issue is whether there's actually going to be another kube-dns
release or not.

miekg:
[ Quoting <notifications@github.com> in "Re: [kubernetes/dns] enable skyDNS ..." ]
>Shuffling seems OK.  I assumed that clients would randomly pick from a
>response set, but maybe they tend to take the first item.  Low cost to
>shuffle, I guess.

No, sadly, most clients this pick the first.

>The real issue is whether there's actually going to be another kube-dns
>release or not.

Doesn't hurt to open issue here: https://github.com/coredns/deployment/issues
and ask for the loadbalance plugin to be enabled.

/Miek

--
Miek Gieben

thockin:
I see zero reason not to do this for CoreDNS (unless it has some cost, but
it seems a random shuffle is fine).

We need to decide if there will be another kube-dns release ever, other
than security...

wu8685:
I have used a forked kube-dns with the RR enabled in my k8s environment. 
It is ok to consider this PR till there is a release.

bboreham:
Came across this by accident and puzzled to see no mention of the "happy eyeballs" algorithm which defeats round-robin via DNS.

https://daniel.haxx.se/blog/2012/01/03/getaddrinfo-with-round-robin-dns-and-happy-eyeballs/

PR #173, opened 2017-11-29
Refactor to use Prometheus Counter instead of Gauge

rramkumar1:
Refactors package sidecar to use Prometheus Counters instead of Gauges. Metrics can be found in pkg/sidecar/metrics.go. Also adds a new test called server_test.go in package sidecar


rramkumar1:
Issue filed

PR #185, opened 2018-01-02
Add validation for upstreamNameserver port

### LINKED ISSUE #178: upstreamNameservers does not allow custom port number
While debugging #169 I found out that `upstreamNameservers` in `kube-dns` configmap does not allow port numbers for upstream servers.

Port numbers are accepted if you give them as `--nameserver=server:port` argument for kube-dns container, but it would be nice to have them accepted via configmap as well.

rramkumar1:
Fixes #178. Add validation for a port number for upstream nameservers specified in the config map.


pkg/dns/config/config.go
```diff
@@ -117,7 +117,15 @@ func (config *Config) validateUpstreamNameserver() error {
 	}
 
 	for _, ns := range config.UpstreamNameservers {
-		if len(validation.IsValidIP(ns)) > 0 {
+		// TODO(rramkumar): Cleanly split this into seperate function.
+		nsStrings := strings.SplitN(ns, ":", 2)
+		// Validate port if specified
+		if len(nsStrings) == 2 {
+			if _, err := strconv.ParseUint(nsStrings[1], 10, 16); err != nil {
+				return fmt.Errorf("Invalid nameserver: %q", ns)
```
m1093782566: nit: first world of error message should be low case.

rramkumar1: I'm not following... what do should the error message be?

miekg:
[ Quoting <notifications@github.com> in "Re: [kubernetes/dns] Add validation..." ]
>rramkumar1 commented on this pull request.
>
>
>
>> @@ -117,7 +117,15 @@ func (config *Config) validateUpstreamNameserver() error {
> 	}
>
> 	for _, ns := range config.UpstreamNameservers {
>-		if len(validation.IsValidIP(ns)) > 0 {
>+		// TODO(rramkumar): Cleanly split this into seperate function.
>+		nsStrings := strings.SplitN(ns, ":", 2)
>+		// Validate port if specified
>+		if len(nsStrings) == 2 {
>+			if _, err := strconv.ParseUint(nsStrings[1], 10, 16); err != nil {
>+				return fmt.Errorf("Invalid nameserver: %q", ns)
>
>I'm not following... what do should the error message be?

~~~
return fmt.Errorf("invalid nameserver: %q", ns)
~~~

Why is net.SplitHostPort not used btw?

rramkumar1:
@miekg I initially thought about SplitHostPort but then I figured it might be cleaner to reuse the validation logic from validateStubDomains(). If you prefer SplitHostPort, I can do that and reimplement the validation in validateStubDomains() to use it.

pkg/dns/config/config.go
m1093782566: fmt.Errorf("invalid nameserver: %q", ns)

glog.Errorf("Invalid ...")

bowei: Generally speaking, if the purpose of a function is to validate something and return an error, you should not need to log as the error will be logged by the caller up on the call stack. This also avoids spuriously error logging when testing potential configurations.

rramkumar1:
@m1093782566 For some reason when I squashed the commits the comment history went away but I addressed the nit. 

@miekg Tried to use SplitHostPort but it errors when given an IP with no port. This is rough because it means there has to be some hacky logic (i.e find last colon, get port manually). I added a TODO to figure out how to do this cleanly. In the meantime, I would at least like to get this in.

pkg/dns/config/config.go
rramkumar1: done

PR #191, opened 2018-01-25
Support DNS A record in the form of "podName.namespace.pod.cluster.local"

langyenan:
This PR assigns a DNS A record for EVERY Pod in the form of "podName.namespace.pod.cluster.local" without needing of service.

**Special notes for your reviewer**:
Before this PR, every pod has a DNS entry based on pod IP, which is kind of inconvenient. (why would I use DNS if I already know other's IP)

**Release note**:
```release-note
assign a DNS A record for EVERY Pod in the form of "podName.namespace.pod.cluster.local" without needing of service.
```

### DIFF
--- a/pkg/dns/dns.go
+++ b/pkg/dns/dns.go
@@ -676,6 +676,9 @@ func (kd *KubeDNS) getRecordsForPath(path []string, exact bool) ([]skymsg.Servic
 	if kd.isPodRecord(path) {
 		ip, err := kd.getPodIP(path)
 		if err == nil {
+			if ip == "" {
+				return []skymsg.Service{}, nil
+			}
 			skyMsg, _ := util.GetSkyMsg(ip, 0)
 			return []skymsg.Service{*skyMsg}, nil
 		}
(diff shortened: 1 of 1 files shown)


thockin:
Philosophical issues aside, I'm very worried about the ability to scale this approach.  If people rely on Pods to have meaningful names, this will get a lot of traffic.  Most pods are expected to live under replicasets, so will have semi-random names anyway.  The obvious technique would be to watch and cache pods, which is very heavy-weight at scale, and we know that some installs run a DNS pod per node.  Watching all pods from all nodes is not viable.

johnbelamaric:
Since `pod` is its own subdomain you can run local instances per-node that do not watch pods, and stub domain `pod` to a central service that does watch pods. That avoids the "every node watches pods issue".

To me, the philosophical issue is maybe more concerning. We do get endpoint names when it's in a StatefulSet w/ a Service today. Adding the pod records like this would encourage specific pods to be "pets" without the abstraction of StatefulSet.

thockin:
Precisely.

langyenan:
Thanks for your replies, the thought you shared is inspiring,  I think this PR is good to be closed.

PR #192, opened 2018-01-29
reduce binary size

oilbeater:
remove symbol tables and debug info to reduce binary size, see https://golang.org/cmd/link/

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -38,5 +38,5 @@ fi
 
 go install                                                         \
     -installsuffix "static"                                        \
-    -ldflags "-X ${PKG}/pkg/version.VERSION=${VERSION}"            \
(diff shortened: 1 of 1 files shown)


bowei:
What is the reduction? Give original size and new size...I'm not sure if this is really necessary.

oilbeater:
dnsmasq-nanny 35M -> 22M
kube-dns 44M-> 28M
sidecar 36M -> 23M

We are trying to build our Kubernetes distribution, and the size is really matter because network condition is not so good everywhere, we want to minimize the network transfer time.

thockin:
Note that if you do this, any stack-dump is going to be missing symbols.  I
don't think we should do this for the upstream binaries.

oilbeater:
@thockin these flags only strip symbols for debugger, other information like panic stack trace still kept

PR #193, opened 2018-02-01
update go version to 1.9.3

oilbeater:
As kubernetes already updated to 1.9.3


thockin:
@bowei I don't know what we're doing with ginkgo here (I wish we would stop using it :) - PTAL when you get a sec

PR #194, opened 2018-02-05
Replace sha1 hash with ~x5 faster spooky hash

Nefelim4ag:
Time for hashing 100000 Int s
sha1:  44.198325ms
spooky:  7.308553ms


thockin:
Thanks for the PR, but we can't just edit vendored code.  You'd have to push these changes upstream and then re-vendor them back down, or else completely fork it into third_party.

I'm not sure I buy that this would have a measurable impact on any real-world workload though.  Do you have a realistic use-case?

Nefelim4ag:
@thockin,
Sorry not found a easy way to get kubedns running and get cache resolution performance.

Issue #196, opened 2018-02-09
difficult to investigate which is problematic on the initialization, services or endpoints


oomichi:
As the issue report of https://github.com/kubernetes/kubernetes/issues/55471 we are facing a problem during the dns initialization and the error log is like:

I0117 00:44:06.332308 1 dns.go:173] Waiting for services and endpoints to be initialized from apiserver...
F0117 00:44:06.832362 1 dns.go:167] Timeout waiting for initialization

The dns is waiting for both services and endpoints to be initialized, but we cannot distinguish which is problematic if facing this log. It would be better to say which cannot be initialized on the log.

PR #197, opened 2018-02-09
Separate the initialization logs for each resource

oomichi:
To investigate issues easily when facing it, this commit separates
the log for each resource, services and endpoints.

fixes: #196

### DIFF
--- a/pkg/dns/dns.go
+++ b/pkg/dns/dns.go
@@ -159,11 +159,19 @@ func (kd *KubeDNS) waitForResourceSyncedOrDie() {
 		case <-timeout:
 			glog.Fatalf("Timeout waiting for initialization")
 		case <-ticker.C:
-			if kd.endpointsController.HasSynced() && kd.serviceController.HasSynced() {
-				glog.V(0).Infof("Initialized services and endpoints from apiserver")
-				return
+			unsyncedResources := []string{}
+			if !kd.endpointsController.HasSynced() {
+				unsyncedResources = append(unsyncedResources, "endpoints")
 			}
-			glog.V(0).Infof("Waiting for services and endpoints to be initialized from apiserver...")
+			if !kd.serviceController.HasSynced() {
+				unsyncedResources = append(unsyncedResources, "services")
+			}
+			if len(unsyncedResources) > 0 {
+				glog.V(0).Infof("Waiting for %v to be initialized from apiserver...", unsyncedResources)
+				continue
+			}
+			glog.V(0).Infof("Initialized services and endpoints from apiserver")
+			return
 		}
 	}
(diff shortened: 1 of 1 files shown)


pkg/dns/dns.go
```diff
@@ -166,11 +166,16 @@ func (kd *KubeDNS) waitForResourceSyncedOrDie() {
...
 		case <-ticker.C:
-			if kd.endpointsController.HasSynced() && kd.serviceController.HasSynced() {
-				glog.V(0).Infof("Initialized services and endpoints from apiserver")
-				return
+			if !kd.endpointsController.HasSynced() {
+				glog.V(0).Infof("Waiting for endpoints to be initialized from apiserver...")
+				continue
+			}
+			if !kd.serviceController.HasSynced() {
+				glog.V(0).Infof("Waiting for services to be initialized from apiserver...")
```
MrHohn: It is also confusing if both endpoints and services fail to be synced while this only says waiting for endpoint. Could you make it check both `endpointsController` and `serviceController` regardless?

oomichi: @MrHohn Thanks for your review, I got it.
I will update this PR soon based on your comment.

pkg/dns/dns.go
```diff
@@ -166,11 +166,22 @@ func (kd *KubeDNS) waitForResourceSyncedOrDie() {
...
+				glog.V(0).Infof("Waiting for endpoints to be initialized from apiserver...")
+				continue
 			}
-			glog.V(0).Infof("Waiting for services and endpoints to be initialized from apiserver...")
+			if !HasSyncedServices {
+				glog.V(0).Infof("Waiting for services to be initialized from apiserver...")
+				continue
+			}
+			glog.V(0).Infof("Initialized services and endpoints from apiserver")
+			return
```
MrHohn: This is a bit odd, what about something like this?
```golang
			unsyncedResources := []string{}
			if !kd.endpointsController.HasSynced() {
				unsyncedResources = append(unsyncedResources, "endpoints")
			}
			if !kd.serviceController.HasSynced() {
				unsyncedResources = append(unsyncedResources, "services")
			}
			if len(unsyncedResources) > 0 {
				glog.V(0).Infof("Waiting for %v to be initialized from apiserver...", unsyncedResources)
				continue
			}
			glog.V(0).Infof("Initialized services and endpoints from apiserver")
			return
```

MrHohn:
FYI travis is currently broken: https://github.com/kubernetes/dns/issues/203.

oomichi:
@MrHohn Thanks for your review and advice.
The issue you pointed has been solved, so I will update this pr soon.

Issue #203, opened 2018-02-27
Travis build is failing on HEAD and PRs


MrHohn:
Ref https://travis-ci.org/kubernetes/dns/builds/346993176:
```
$ cd ${GOPATH}/src/k8s.io/dns && bin/amd64/ginkgo test/e2e
/home/travis/.travis/job_stages: line 57: bin/amd64/ginkgo: No such file or directory

The command "cd ${GOPATH}/src/k8s.io/dns && bin/amd64/ginkgo test/e2e" exited with 127.
```

/assign

thockin:
I emailed Bowei about this - the build is all messed up

PR #206, opened 2018-03-02
Fix travis: include vendor folder in build.sh

MrHohn:
Did a bit digging and found the CI build was broken after bumping to Go 1.9+ (https://github.com/kubernetes/dns/pull/193).

From Go 1.9 release notes, looks like command `go install ./...` will no longer matches package in vendor folder (https://golang.org/doc/go1.9#vendor-dotdotdot), which breaks our CI build because we are expecting `bin/amd64/ginkgo`.

This PR adds the vendor folder back for installing. Might not be the best choice but I think it at least works.

Fix https://github.com/kubernetes/dns/issues/203.

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -36,7 +36,13 @@ if [ $GOARCH == "amd64" ]; then
     export GOBIN="$GOPATH/bin/linux_amd64"
 fi
 
+# ./vendor/... is specified to include installing binary from vendor folder.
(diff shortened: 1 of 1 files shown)


MrHohn:
@thockin @bowei

thockin:
Can you add a comment about why?

MrHohn:
> Can you add a comment about why?

Sure, added some comments, hope that makes sense.

PR #214, opened 2018-03-20
Delete docker pull from Makefiles

Nefelim4ag:
That fixув building on closed clusters,
where for security reasons no internet connection are available
and needed images are unpacked to build machine

Docker will pull missing images self

For dnsmasq --pull useless because specific version of image
are set in file, so no new images will come.

Signed-off-by: Timofey Titovets <nefelim4ag@gmail.com>

### DIFF
--- a/images/dnsmasq/Makefile
+++ b/images/dnsmasq/Makefile
@@ -75,6 +75,8 @@ BUILDSTAMP := $(subst :,_,$(subst /,_,$(REGISTRY))_$(IMAGE)_$(VERSION))
 CONTAINER_STAMP := .$(BUILDSTAMP)-container
 PUSH_STAMP := .$(BUILDSTAMP)-push
 
+DOCKER_PULL ?= 1
+
 ifeq ($(VERBOSE), 1)
 	VERBOSE_OUTPUT := >&1
 else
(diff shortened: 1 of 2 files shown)


Nefelim4ag:
P.S.
If you found that meaning, cherry pick please to 1.14.10

Thanks.

bowei:
I think this should be an option to build, not remove pull entirely.

Nefelim4ag:
@bowei, option like CONTAINER_BINARIES=kube-dns ?
Add something like DOCKER_PULL=disable ?

```
make DOCKER_PULL=disable
```

bowei:
yes, that would be ok

Nefelim4ag:
@bowei, PR updated, now:
```
make DOCKER_PULL=0
```
similar to:
```
make VERBOSE=1
```

idealhack:
Do we need this change?

/ok-to-test

Nefelim4ag:
Because not everyone have a access to external sources while build.

ixdy:
Docker Hub reuses tags (e.g. `alpine:3.6` today is probably not the same `alpine:3.6` from a month ago), so we wanted to always pull the base image to ensure we weren't using an old cached version. (This had been a problem when I added the `--pull`.)

I'm fine if pulling is enabled by default but can be turned off when folks know what they're doing (e.g. building in an airgapped environment).

(The long-term correct solution would be to stop depending on tags which are mutable, i.e. use digests instead.)

PR #215, opened 2018-03-20
Implement HasSynced() of skydns Backend interface

### LINKED ISSUE #186: kube-dns returns NXDOMAIN when not yet synced with apiserver
From https://github.com/kubernetes/kubernetes/issues/57111

**What happened**:
when kube-dns has not yet synced with the apiserver, it return NXDOMAIN for all cluster names

**What you expected to happen**:
It should return SERVFAIL

**Anything else we need to know?**:
NXDOMAIN is the wrong status. It even can be cached by downstream DNS servers.

**Environment**:
- Kubernetes version (use `kubectl version`): 1.8.4

@bowei @rramkumar1 @johnbelamaric

rramkumar1:
This PR fixes #186. 

We implement the HasSynced() interface that skydns defines. In the case that the endpoints and services have not been initally synced from the API server, skydns will now refuse the connection instead of returning NXDOMAIN.


MrHohn:
Ref the skydns upstream PR: https://github.com/skynetservices/skydns/pull/327.

bowei:
You didn't vendor everything. It doesn't seem to build.

rramkumar1:
Trying again in a separate PR...

PR #222, opened 2018-03-31
Update cache.go

YEXINGZHE54:
currently, EvictRandom will clean up all cached entries when cache is full. fix it, only clean additional ones


YEXINGZHE54:
yes, it's a correct way.

Issue #224, opened 2018-04-13
DNS resolution for externalName services broken in v1.14.9


ApsOps:
After updating to v1.14.9, externalName services are not resolving anymore. I tested with v1.14.8 and it works.

```
/ # dig @100.102.0.17 mysql-external.default.svc.cluster.local

; <<>> DiG 9.11.2-P1 <<>> @100.102.0.17 mysql-external.default.svc.cluster.local
; (1 server found)
;; global options: +cmd
;; Got answer:
;; WARNING: .local is reserved for Multicast DNS
;; You are currently testing what happens when an mDNS query is leaked to DNS
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 45659
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 0

;; QUESTION SECTION:
;mysql-external.default.svc.cluster.local.	IN A

;; AUTHORITY SECTION:
cluster.local.		60	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 1523602800 28800 7200 604800 60

;; Query time: 1 msec
;; SERVER: 100.102.0.17#53(100.102.0.17)
;; WHEN: Fri Apr 13 07:58:15 UTC 2018
;; MSG SIZE  rcvd: 110
```

kube-dns pod logs (I've replaced the correct domain with example.com but it exists and resolves normally in v1.14.8):

```
[kube-dns-aps-78b75f775c-b2j8p kubedns] I0413 08:12:22.197655       1 dns.go:612] Query for "mysql-external.default.svc.cluster.local.", exact: false
[kube-dns-aps-78b75f775c-b2j8p kubedns] I0413 08:12:22.197682       1 dns.go:841] Not a federation query: len(["mysql-external" "default" "svc" "cluster" "local"]) != 4+len(["local" "cluster"])
[kube-dns-aps-78b75f775c-b2j8p kubedns] I0413 08:12:22.197700       1 dns.go:732] Found 1 records for [local cluster svc default mysql-external] in the cache
[...]

PR #225, opened 2018-04-14
Fix external name not solving by reloading resolv.conf.

grayluck:
Now skydns will be configured with resolv.conf when nameserver is empty.

Fixes#224 (https://github.com/kubernetes/dns/issues/224)

\assign @MrHohn

### DIFF
--- a/pkg/dns/dns.go
+++ b/pkg/dns/dns.go
@@ -19,7 +19,6 @@ package dns
 import (
 	"fmt"
 	"net"
-	"os"
 	"strings"
 	"sync"
 	"time"
@@ -52,6 +51,10 @@ const (
 
 	// Resync period for the kube controller loop.
 	resyncPeriod = 5 * time.Minute
+)
+
+var (
+	defaultResolvFile = "/etc/resolv.conf"
 )
 
 type KubeDNS struct {
@@ -141,16 +144,16 @@ func NewKubeDNS(client clientset.Interface, clusterDomain string, timeout time.D
 }
 
 func (kd *KubeDNS) loadDefaultNameserver() []string {
-	nameservers := []string{}
-	if c, err := dns.ClientConfigFromFile("/etc/resolv.conf"); !os.IsNotExist(err) {
-		if err != nil {
-			return nameservers
-		}
+	if c, err := dns.ClientConfigFromFile(defaultResolvFile); err != nil {
+		glog.Errorf("Load nameserver from resolv.conf failed: %v", err)
+		return []string{}
+	} else {
+		nameservers := []string{}
 		for _, s := range c.Servers {
 			nameservers = append(nameservers, net.JoinHostPort(s, c.Port))
 		}
+		return nameservers
 	}
-	return nameservers
 }
 
 func (kd *KubeDNS) updateConfig(nextConfig *config.Config) {
@@ -161,7 +164,11 @@ func (kd *KubeDNS) updateConfig(nextConfig *config.Config) {
 		var nameServers []string
 		for _, nameServer := range nextConfig.UpstreamNameservers {
 			if ip, port, err := util.ValidateNameserverIpAndPort(nameServer); err != nil {
-				glog.V(1).Infof("Invalid nameserver %q: %v", nameServer, err)
+				glog.Errorf("Invalid nameserver %q: %v", nameServer, err)
+				if len(kd.SkyDNSConfig.Nameservers) == 0 {
+					// Fall back to resolv.conf on initialization failed.
+					kd.SkyDNSConfig.Nameservers = kd.loadDefaultNameserver()
+				}
 				return
 			} else {
 				nameServers = append(nameServers, net.JoinHostPort(ip, port))
--- a/pkg/dns/dns.go
+++ b/pkg/dns/dns.go
@@ -166,7 +166,7 @@ func (kd *KubeDNS) updateConfig(nextConfig *config.Config) {
 			if ip, port, err := util.ValidateNameserverIpAndPort(nameServer); err != nil {
 				glog.Errorf("Invalid nameserver %q: %v", nameServer, err)
 				if len(kd.SkyDNSConfig.Nameservers) == 0 {
-					// Fall back to resolv.conf on initialization failed.
+					// Fall back to resolv.conf on initialization failure.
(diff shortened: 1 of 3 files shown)


pkg/dns/dns.go
```diff
@@ -153,7 +167,9 @@ func (kd *KubeDNS) updateConfig(nextConfig *config.Config) {
 				nameServers = append(nameServers, net.JoinHostPort(ip, port))
 			}
 		}
-		kd.SkyDNSConfig.Nameservers = nameServers
+		if len(nameServers) == 0 {
+			kd.SkyDNSConfig.Nameservers = kd.loadDefaultNameserver()
+		}
```
MrHohn: Seems like `kd.SkyDNSConfig.Nameservers = nameServers` is still needed?

grayluck: Fixed..

MrHohn:
Keep @bowei in the loop.

pkg/dns/dns.go
```diff
@@ -148,12 +164,20 @@ func (kd *KubeDNS) updateConfig(nextConfig *config.Config) {
 		for _, nameServer := range nextConfig.UpstreamNameservers {
 			if ip, port, err := util.ValidateNameserverIpAndPort(nameServer); err != nil {
 				glog.V(1).Infof("Invalid nameserver %q: %v", nameServer, err)
```
MrHohn: glog.Errorf()?

pkg/dns/dns.go
```diff
@@ -139,6 +143,18 @@ func NewKubeDNS(client clientset.Interface, clusterDomain string, timeout time.D
...
 }
 
+func (kd *KubeDNS) loadDefaultNameserver() []string {
+	nameservers := []string{}
+	if c, err := dns.ClientConfigFromFile(defaultResolvFile); err == nil {
+		for _, s := range c.Servers {
+			nameservers = append(nameservers, net.JoinHostPort(s, c.Port))
+		}
+	} else {
+		glog.V(1).Infof("Load nameserver from resolv.conf failed: %v", err)
```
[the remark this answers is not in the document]
grayluck: Done.

pkg/dns/dns.go
grayluck: Done.

pkg/dns/dns.go
```diff
@@ -139,6 +143,19 @@ func NewKubeDNS(client clientset.Interface, clusterDomain string, timeout time.D
 	return kd
 }
 
+func (kd *KubeDNS) loadDefaultNameserver() []string {
+	if c, err := dns.ClientConfigFromFile(defaultResolvFile); err != nil {
+		glog.Errorf("Load nameserver from resolv.conf failed: %v", err)
+		return []string{}
+	} else {
```
MrHohn: The else block seems unnecessary.

MrHohn: I see, we are using the`c`.

MrHohn: No worries, let keep this as is.

pkg/dns/dns.go
```diff
@@ -147,13 +164,21 @@ func (kd *KubeDNS) updateConfig(nextConfig *config.Config) {
 		var nameServers []string
 		for _, nameServer := range nextConfig.UpstreamNameservers {
 			if ip, port, err := util.ValidateNameserverIpAndPort(nameServer); err != nil {
-				glog.V(1).Infof("Invalid nameserver %q: %v", nameServer, err)
+				glog.Errorf("Invalid nameserver %q: %v", nameServer, err)
+				if len(kd.SkyDNSConfig.Nameservers) == 0 {
+					// Fall back to resolv.conf on initialization failed.
+					kd.SkyDNSConfig.Nameservers = kd.loadDefaultNameserver()
```
MrHohn: I think we should fall through here --- append whatever valid nameservers are passed in, and reuse the 0 length detection logic below:
```golang
if ip, port, err := util.ValidateNameserverIpAndPort(nameServer); err != nil {
   glog.Errorf("Invalid nameserver %q: %v", nameServer, err)
} else {
   nameServers = append(nameServers, net.JoinHostPort(ip, port))
}
...
```

grayluck: The idea is, we don't want to pollute kd.config with invalid input. (But the truth is, kd.config.UpstreamNameservers are never used so it's okay to be polluted)

MrHohn: Sounds good, though having duplicated logic here seems bad. What about:
```golang
	if kd.SkyDNSConfig != nil {
		var nameServers []string
		for _, nameServer := range nextConfig.UpstreamNameservers {
			if ip, port, err := util.ValidateNameserverIpAndPort(nameServer); err != nil {
				glog.Errorf("Invalid nameserver %q: %v", nameServer, err)
				// Don't populate nameServers if invalid input is detected.
				nameServers = []string{}
				break
			} else {
				nameServers = append(nameServers, net.JoinHostPort(ip, port))
			}
		}
		if len(nameServers) == 0 {
			...
```

MrHohn: Sorry what I wrote is wrong. I got what you were saying.

grayluck: But we can still do the partially update and leave kd.config untouched. Do you want me to do that?
I am kind of against this since kd.SkyDNSConfig is partially updated and kd.config is not which might be confusing in some extreme cases.

pkg/dns/dns.go
```diff
@@ -147,13 +164,21 @@ func (kd *KubeDNS) updateConfig(nextConfig *config.Config) {
 		var nameServers []string
 		for _, nameServer := range nextConfig.UpstreamNameservers {
 			if ip, port, err := util.ValidateNameserverIpAndPort(nameServer); err != nil {
-				glog.V(1).Infof("Invalid nameserver %q: %v", nameServer, err)
+				glog.Errorf("Invalid nameserver %q: %v", nameServer, err)
+				if len(kd.SkyDNSConfig.Nameservers) == 0 {
+					// Fall back to resolv.conf on initialization failed.
```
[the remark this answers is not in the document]
grayluck: Fixed.

grayluck:
Thanks for review!

MrHohn:
Will merge this once travis passes.

Issue #234, opened 2018-05-18
response for a headless service with no ready endpoints


chrisohaver:
kube-dns replies with NXDOMAIN, for a headless service that has no ready endpoints.
But the [spec](https://github.com/kubernetes/dns/blob/master/docs/specification.md#241---a-records) says that the response should be NODATA, "because the service exists".

Presumably the spec wrong, since the spec was originally based off of kube-dns behavior at the time.
Or, in this case, is the spec behavior preferred?

miekg:
[ Quoting <notifications@github.com> in "[kubernetes/dns] response for a hea..." ]
>kube-dns replies with NXDOMAIN, for a headless service that has no ready endpoints.
>But the [spec](https://github.com/kubernetes/dns/blob/master/docs/specification.md#241---a-records) says that the response should be NODATA, "because the service exists".

I take it 'service exist' means the domain name exist, but there is no rdata; a 
NODATA response is correct in that case.

chrisohaver:
> I take it 'service exist' means the domain name exist...

In the context of the spec, "service exists" only means that there is a headless service defined in kubernetes.  Whether or not the cluster DNS should produce a domain with that service name is the question.

> NODATA response is correct in that case.

On the other hand, for a headless service with no ready endpoints, there is no record at that level, or beneath it.  Which suggests NXDOMAIN is more correct. e.g. 
`headless.default.svc.cluster.local` -> no records at this level
`*. headless.default.svc.cluster.local` -> no records beneath it

miekg:
[ Quoting <notifications@github.com> in "Re: [kubernetes/dns] response for a..." ]
>> I take it 'service exist' means the domain name exist...
>
>In the context of the spec, "service exists" only means that there is a headless service defined in kubernetes.  Whether or not the cluster DNS should produce a domain with that service name is the question.
>
>> NODATA response is correct in that case.
>
>On the other hand, for a headless service with no ready endpoints, there is no record at that level, or beneath it.  Which suggests NXDOMAIN is more correct. e.g.
>`headless.default.svc.cluster.local` -> no records at this level
>`*. headless.default.svc.cluster.local` -> no records beneath it

fair enough, but nodata/nxdomain should matter to a client.

chrisohaver:
> nodata/nxdomain should matter to a client.

Yes (e.g. search path completion), which is why changing the current implementation to match the spec should be taken delicately.  Changing the spec to match the implementation has less disruptive potential.

[...]

PR #245, opened 2018-06-12
DNS Spec: Headless service with no ready endpoints 

chrisohaver:
Proposing that a headless service without any ready endpoints should return `NXDOMAIN`, to match current behavior of kube-dns and CoreDNS.

I'm proposing we update the spec to match the implementation of kube-dns, because the spec was initially based on kube-dns behavior.

If we don't update the spec, then we should open bugs for kube-dns (and CoreDNS) and fix.

see #234

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -104,10 +104,9 @@ Given a headless Service `<service>` in Namespace `<ns>` (i.e., a Service with
 no ClusterIP), the following records must exist.
 
 #### 2.4.1 - `A` Records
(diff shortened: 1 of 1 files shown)


chrisohaver:
thoughts?

johnbelamaric:
I agree it makes sense to update the spec to match the current behavior. Changing the behavior could have unexpected consequences for users.

chrisohaver:
How can we move this forward.

PR #251, opened 2018-08-08
Bump build image to go 1.10.3

cmluciano:
Signed-off-by: Christopher M. Luciano <cmluciano@us.ibm.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ REGISTRY ?= staging-k8s.gcr.io
 # Default architecture to build for.
 ARCH ?= amd64
 # Image to use for building.
-BUILD_IMAGE ?= golang:1.9.3-alpine
+BUILD_IMAGE ?= golang:1.10.3-alpine
 # Containers will be named: $(CONTAINER_PREFIX)-$(BINARY)-$(ARCH):$(VERSION)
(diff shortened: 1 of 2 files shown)


MrHohn:
Seems like `go vet` is failing after this bump (https://travis-ci.org/kubernetes/dns/builds/413763836?utm_source=github_status&utm_medium=notification):
```
Checking go vet: FAIL
go: disabling cache (/.cache/go-build) due to initialization failure: mkdir /.cache: permission denied
make: *** [test] Error 1
```

Dig a bit and found https://github.com/docker-library/golang/issues/225, docker sets `HOME='/'` if container runs as non-root user. Go cache is not disabled in this case hence emits above warning message. This is fixed upstream (https://github.com/golang/go/issues/26280) but looks like the milestone is 1.12.

We can probably just set `HOME=''` so that caching will be disabled. Specifically change https://github.com/kubernetes/dns/blob/b333dce3623db797e3998c3cc383384019200afc/build/test.sh#L43 to:
```shell
ERRS=$(HOME='' go vet ${TARGETS} 2>&1 || true)
```

cmluciano:
Wow that is an interesting chain of events; nice find. This would only disable caching for the go vet command right?

PR #258, opened 2018-09-10
Fix DNS query to get metrics from Dnsmasq

pedrosland:
This PR fixes the DNS queries to get metrics such as `hits.bind.`

Without this change, the sidecar doesn't report `kubedns_dnsmasq_hits` or similar metrics and it produces logs like:

```
W0910 16:07:40.684579   18427 server.go:64] Error getting metrics from dnsmasq: invalid number of Answer records for hits.bind.: 0
```

It seems that Dnsmasq requires the recursive flag to respond to these queries. Using dig confirms this. 

Working:
```
dig chaos txt hits.bind.
```
Not working:
```
dig +norecurse chaos txt hits.bind.
```

### DIFF
--- a/pkg/dnsmasq/metrics.go
+++ b/pkg/dnsmasq/metrics.go
@@ -91,7 +91,7 @@ func (mc *metricsClient) GetMetrics() (ret *Metrics, err error) {
 func (mc *metricsClient) getSingleMetric(name string) (int64, error) {
 	msg := new(dns.Msg)
 	msg.Id = dns.Id()
-	msg.RecursionDesired = false
+	msg.RecursionDesired = true
 	msg.Question = make([]dns.Question, 1)
 	msg.Question[0] = dns.Question{
 		Name:   name,


MrHohn:
Thanks for fixing this, I can see the same errors being emitted on my dev cluster as well:
```
...
W1006 01:01:52.601651       1 server.go:64] Error getting metrics from dnsmasq: invalid number of Answer records for hits.bind.: 0
I1006 01:01:53.764179       1 dnsprobe.go:118] Sending DNS request @127.0.0.1:10053 kubernetes.default.svc.cluster.local.
I1006 01:01:53.765771       1 dnsprobe.go:120] Got response, err=<nil> after 1.160682ms
I1006 01:01:53.765852       1 dnsprobe.go:56] Sleeping 4.998839318s
I1006 01:01:55.806292       1 dnsprobe.go:118] Sending DNS request @127.0.0.1:53 kubernetes.default.svc.cluster.local.
I1006 01:01:55.807492       1 dnsprobe.go:120] Got response, err=<nil> after 947.979µs
I1006 01:01:55.807536       1 dnsprobe.go:56] Sleeping 4.999052021s
W1006 01:01:57.602561       1 server.go:64] Error getting metrics from dnsmasq: invalid number of Answer records for hits.bind.: 0
I1006 01:01:58.765267       1 dnsprobe.go:118] Sending DNS request @127.0.0.1:10053 kubernetes.default.svc.cluster.local.
I1006 01:01:58.766349       1 dnsprobe.go:120] Got response, err=<nil> after 869.163µs
I1006 01:01:58.766428       1 dnsprobe.go:56] Sleeping 4.999130837s
```

Wonder would it be related to a recent change in dnsmasq 2.79 (http://www.thekelleys.org.uk/dnsmasq/CHANGELOG):
```
	Always return a SERVFAIL answer to DNS queries without the
	recursion desired bit set, UNLESS acting as an authoritative
	DNS server. This avoids a potential route to cache snooping.
```

Though we are currently using dnsmasq 2.78 so I'm not sure :/ https://github.com/kubernetes/dns/blob/1.14.13/images/dnsmasq/Makefile#L18

cc @prameshj

PR #262, opened 2018-09-19
update dnsmasq to 2.79

matt-potter:
Specifically this provides min-port at a sane default (1024), which would aid our deployments with kube-aws, where are are unable to pass custom flags to the dnsmasq process

### DIFF
--- a/images/dnsmasq/Makefile
+++ b/images/dnsmasq/Makefile
@@ -15,7 +15,7 @@
 VERSION ?= $(shell git describe --tags --always --dirty)
 REGISTRY ?= staging-k8s.gcr.io
 ARCH ?= amd64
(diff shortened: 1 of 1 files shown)


bowei:
cc: @prameshj

prameshj:
The e2e tests in kubernetes/kubernetes need to be fixed before upgrading to dnsmasq 2.79
https://github.com/kubernetes/kubernetes/issues/69123

bowei:
@prameshj is this still relevant?

prameshj:
/remove-lifecycle rotten
/ok-to-test

The kubernetes e2e tests have been fixed.
@matt-potter Do we still need this change?

PR #267, opened 2018-09-27
Purge the manifest after the push

mkumatag:
Triggered by https://github.com/kubernetes/kubernetes/issues/69104#issuecomment-424799217


ixdy:
Thanks!

PR #127, opened 2017-07-17
Use alpine as base image on all platforms

MrHohn:
Ack https://github.com/kubernetes/contrib/pull/2670#issuecomment-315712954.

@guirish Would you be able to test the s390x image? Thanks.

/assign @bowei 

cc @GheRivero @gajju26

PR #268, opened 2018-10-01
Remove unused NOBODY/ARG_NOBODY arg from docker build

ixdy:
`NOBODY` and `ARG_NOBODY` were added in #31 because amd64 and the other architectures were using different base images, but its need was basically but removed in #127 when all architectures were converted to use alpine.

The names were also confusing (they apply to the group, not the user), so let's just remove them and hardcode the group in the Dockerfile again.


MrHohn:
Not sure why test job failed to report back the status, from the link I can tell it succeeded: https://travis-ci.org/kubernetes/dns/builds/435758895.

Manually merging this.

PR #269, opened 2018-10-30
Migrate from Godep to dep for managing dependencies.

prameshj:


MrHohn:
We might need to update https://github.com/kubernetes/dns/blob/master/docs/go-dependencies.md accordingly. But that can happen later.

prameshj:
I have the doc change here - https://github.com/prameshj/dns/blob/5220146ca76fb49609626579e6b64372a0c096d4/docs/go-dependencies.md
Will pull it into this branch when we are closer to merge.

PR #274, opened 2018-11-11
Upgrade CoreDNS to 1.2.6

prameshj:
this contains a fix in the loop plugin that was incorrectly detecting a loop when upstream dns was unavailable.
coredns/coredns#2255


justinsb:
/ok-to-test

/lgtm

I scanned over the vendor changes and they look reasonable also (and include the fix we really want)

PR #279, opened 2018-12-02
allow service name for stubDomains

### LINKED ISSUE #82: kube-dns doesn't support using service name in stubDomains as docs and code seem to suggest
According to the [announcement](http://blog.kubernetes.io/2017/04/configuring-private-dns-zones-upstream-nameservers-kubernetes.html) blog post, I should be able to use a service as an upstream for stubDomains (note the note):

```
stubDomains (optional)

    Format: a JSON map using a DNS suffix key (e.g.; “acme.local”) and a value consisting of a JSON array of DNS IPs.
    Note: The target nameserver may itself be a Kubernetes service. For instance, you can run your own copy of dnsmasq to export custom DNS names into the ClusterDNS namespace.
```

However if I try to point at our unbound service, the dnsmasq container crashes:

```
I0505 21:26:49.795586       1 main.go:76] opts: {{/usr/sbin/dnsmasq [-k --cache-size=1000 --log-facility=- --server=/cluster.local/127.0.0.1#10053 --server=/in-addr.arpa/127.0.0.1#10053 --server=/ip6.arpa/127.0.0.1#10053] true} /etc/k8s/dns/dnsmasq-nanny 10000000000}
I0505 21:26:49.795988       1 sync.go:167] Updated stubDomains to map[example.com:[unbound]]
I0505 21:26:49.796066       1 nanny.go:86] Starting dnsmasq [-k --cache-size=1000 --log-facility=- --server=/cluster.local/127.0.0.1#10053 --server=/in-addr.arpa/127.0.0.1#10053 --server=/ip6.arpa/127.0.0.1#10053 --server /example.com/unbound]
I0505 21:26:49.797178       1 nanny.go:111]
W0505 21:26:49.797301       1 nanny.go:112] Got EOF from stdout
F0505 21:26:49.797341       1 nanny.go:182] dnsmasq exited: exit status 1
I0505 21:26:49.797283       1 nanny.go:115]
E0505 21:26:49.797501       1 nanny.go:116] Error reading from stderr: read |0: bad file descriptor
```

[This test](https://github.com/kubernetes/dns/blob/1.14.1/pkg/dns/config/config_test.go#L32) also seems to indicate that names are allowed in stubDomains, but from what I can tell, dnsmasq doesn't allow a name in a `--server` parameter (in the above example, it's sending `--server /example.com/unbound`)

We can currently use the clusterIP of a service as upstream DNS, but that seems pretty error prone.  Is there something else we should be doing here, or am I misinterpreting the documentation?

andy2046:
Fixes #82 

it should work for not only kubernetes service name but also private dns record (e.g. dns.my.domain defined in GKE private dns)

### DIFF
--- a/pkg/dnsmasq/nanny.go
+++ b/pkg/dnsmasq/nanny.go
@@ -20,6 +20,7 @@ import (
 	"bufio"
 	"fmt"
 	"io"
+	"net"
 	"os/exec"
 	"strings"
 
@@ -69,6 +70,11 @@ func (n *Nanny) Configure(args []string, config *config.Config) {
 
 	for domain, serverList := range config.StubDomains {
 		for _, server := range serverList {
+			if isIP := (net.ParseIP(server) != nil); !isIP {
+				if IPs, err := net.LookupIP(server); err == nil && len(IPs) > 0 {
+					server = IPs[0].String()
+				}
+			}
 			// dnsmasq port separator is '#' for some reason.
 			server = munge(server)
 			n.args = append(
--- a/pkg/dnsmasq/nanny.go
+++ b/pkg/dnsmasq/nanny.go
@@ -18,6 +18,7 @@ package dnsmasq
 
 import (
 	"bufio"
+	"context"
 	"fmt"
 	"io"
 	"net"
@@ -53,7 +54,7 @@ func ExtractDnsmasqArgs(cmdlineArgs *[]string) []string {
 }
 
 // Configure the nanny. This must be called before Start().
-func (n *Nanny) Configure(args []string, config *config.Config) {
+func (n *Nanny) Configure(args []string, config *config.Config, kubednsServer string) {
 	n.args = args
 
 	munge := func(s string) string {
@@ -69,10 +70,24 @@ func (n *Nanny) Configure(args []string, config *config.Config) {
 	}
 
 	for domain, serverList := range config.StubDomains {
+		resolver := &net.Resolver{
+			PreferGo: true,
+			Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
+				d := net.Dialer{}
+				return d.DialContext(ctx, "udp", kubednsServer)
+			},
+		}
 		for _, server := range serverList {
 			if isIP := (net.ParseIP(server) != nil); !isIP {
-				if IPs, err := net.LookupIP(server); err == nil && len(IPs) > 0 {
-					server = IPs[0].String()
+				switch {
+				case strings.HasSuffix(server, "cluster.local"):
+					if IPs, err := resolver.LookupIPAddr(context.Background(), server); err == nil && len(IPs) > 0 {
+						server = IPs[0].String()
+					}
+				default:
+					if IPs, err := net.LookupIP(server); err == nil && len(IPs) > 0 {
+						server = IPs[0].String()
+					}
 				}
 			}
 			// dnsmasq port separator is '#' for some reason.
@@ -172,7 +187,7 @@ type RunNannyOpts struct {
 }
 
 // RunNanny runs the nanny and handles configuration updates.
-func RunNanny(sync config.Sync, opts RunNannyOpts) {
+func RunNanny(sync config.Sync, opts RunNannyOpts, kubednsServer string) {
 	defer glog.Flush()
 
 	currentConfig, err := sync.Once()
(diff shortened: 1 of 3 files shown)


andy2046:
I signed it

pkg/dnsmasq/nanny.go
```diff
@@ -69,6 +70,11 @@ func (n *Nanny) Configure(args []string, config *config.Config) {
 
 	for domain, serverList := range config.StubDomains {
 		for _, server := range serverList {
+			if isIP := (net.ParseIP(server) != nil); !isIP {
+				if IPs, err := net.LookupIP(server); err == nil && len(IPs) > 0 {
```
MrHohn: Does this lookup work for the initial setup --- when dnsmasq has not been started?
https://github.com/kubernetes/dns/blob/a038db946ea2aeefe848116325c96644915f503d/pkg/dnsmasq/nanny.go#L179-L182

chrisohaver:
> One issue, the lookup is done at start up, so, if the IP of the hostname changes, the stub-domain will stop resolving.

Actually - i see that it synchronizes periodically, so it would only stop working until next resync.

andy2046:
@chrisohaver Chris thanks for the comments,

- yes, there is a periodic sync of ConfigMap, is it necessary to add another sync of `service` `endpoint` change and restart `dnsmasq` if IP changes?

- for `cluster.local`, one way is to get records from `/cache` endpoint I mentioned above to resolve it, any better suggestions are welcome

thanks again for all the inputs

andy2046:
since `dnsmasq-nanny` relies on `kubedns` for kubernetes cluster local DNS records, I can wait for `kubedns` `healthzPort` before trying to resolve the service name

andy2046:
@MrHohn Zihong, I updated the PR, now it should be able to resolve `cluster.local` SRV records, pls help to review

andy2046:
@chrisohaver Chris FYI, now I use `kubedns` server to resolve `cluster.local` SRV records

pkg/dnsmasq/nanny.go
```diff
@@ -68,7 +70,26 @@ func (n *Nanny) Configure(args []string, config *config.Config) {
...
+			PreferGo: true,
+			Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
+				d := net.Dialer{}
+				return d.DialContext(ctx, "udp", kubednsServer)
+			},
+		}
 		for _, server := range serverList {
+			if isIP := (net.ParseIP(server) != nil); !isIP {
+				switch {
+				case strings.HasSuffix(server, "cluster.local"):
```
MrHohn: Not necessarily in this PR, but we need to ensure this case is covered in e2e test: https://github.com/kubernetes/kubernetes/blob/master/test/e2e/network/dns_configmap.go.

andy2046: @MrHohn Zihong, I will raise another PR in kubernetes repo to cover this case in e2e test

andy2046:
@MrHohn Zihong, I updated the PR based on your comments, please let me know if there is any further changes required, thx

MrHohn:
Thanks! Would defer to @bowei for a final look.

pkg/dnsmasq/nanny.go
```diff
@@ -68,7 +70,30 @@ func (n *Nanny) Configure(args []string, config *config.Config) {
...
+					if IPs, err := resolver.LookupIPAddr(context.Background(), server); err != nil {
+						glog.Errorf("Error looking up IP for %s: %v", server, err)
+					} else if len(IPs) > 0 {
```
bowei: if len(IPs) == 0 {
  glog.Errorf("Name %q does not resolve to any IPs", server)
}

add to the default as well.

pkg/dnsmasq/nanny.go
```diff
@@ -52,7 +54,7 @@ func ExtractDnsmasqArgs(cmdlineArgs *[]string) []string {
 }
 
 // Configure the nanny. This must be called before Start().
-func (n *Nanny) Configure(args []string, config *config.Config) {
+func (n *Nanny) Configure(args []string, config *config.Config, kubednsServer string) {
```
bowei: add to the comment
```
//  Configure .... kubednsServer is the address of the local kubedns instance used to do name resolution for non-IP names.
```

PR #280, opened 2018-12-06
Create script to run node-local-dns cache without kubelet support

justinsb:
This enables node-local-dns to be used experimentally prior to kubelet support.

### DIFF
--- a/tools/node-cache/README.md
+++ b/tools/node-cache/README.md
@@ -1,0 +1,92 @@
+## Overview
+
+This example allows for use of the [node-local-dns
+cache](https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/dns/nodelocaldns),
+without requiring the kubelet `--cluster-dns` flag to be altered.  In
+particular, this means it can be used on GKE clusters.
+
+This is experimental, but feedback is welcome - please tag @justinsb.
+
+## Warning: Network policy
+
+If running with network policy, please see the [README for
+node-local-dns](https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/dns/nodelocaldns#network-policy-and-dns-connectivity);
+network policy will likely need to be configured for the node-local-dns agent.
+
+## Other caveats
+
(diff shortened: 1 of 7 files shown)


tools/node-cache/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Removal
+
+Removal is more complicated that installation.  We can remove the daemonset, and as
+part of pod shutdown the node-local-dns cache should remove the IP interception
+rules.  However, if something goes wrong with the removal, the IP interception rules
+will remain in place, but the node-local-dns cache will not be running to serve
+the intercepted traffic, and DNS lookup will be broken on that node.  However,
+restarting the machine will remove the IP interception rules, so if this is done
+as part of a cluster update the system will self-heal.
```
krmayankk: Is this the only reason why we would not make this the normal way of using dns local cache and instead require explicitly changing kubelet --cluster-dns option ?

krmayankk: I feel like trying this out, quickly since we are running into this a lot. What other drawbacks we can think of of this approach ?

krmayankk:
Must say, this is great work @justaugustus :-)

tools/node-cache/README.md
prameshj: The main reason to modify --cluster-dns and use a new ip was because this approach of re-using service Ip does not work very well in ipvs mode of kube-proxy. In ipvs mode, kube-proxy creates a local interface on the host binding the service ip there. So packets destined to service ip will hit ipvs rules and get forwarded to pods. If ipvs mode does not apply in the setup, this approach will work well.

krmayankk: i think we should include this information about its limitation in ipvs mode in a limitations/caveats section to inform users using that mode.

aledbf:
Any reason why this is not merged?
Thanks

prameshj:
> Any reason why this is not merged?
> Thanks

Since this is experimental and there are a few outstanding comments.

Issue #282, opened 2018-12-21
Node local DNS creates dummy interface without IP address


negz:
Hello,

I've just started experimenting with the new node local DNS cache. For reasons I haven't yet determined the `NetIfManager` manages to create the `nodelocaldns` dummy IP in my setup, but fails to allocate it an IP address. `NetIfManager` [does not check the error return value](https://github.com/kubernetes/dns/blob/401314d/pkg/netif/netif.go#L26) when assigning IPs, so this manifests as follows:

```bash
$ kubectl --kubeconfig=/Users/negz/tfk-negz.kubecfg -n kube-system logs nodelocaldns-ltlfp --previous
2018/12/21 03:22:52 2018-12-21T03:22:52.066Z [INFO] Tearing down
2018/12/21 03:22:53 2018-12-21T03:22:53.064Z [INFO] Setting up networking for node cache
listen tcp 169.254.20.10:8080: bind: cannot assign requested address
```

I see when inspect the dummy interface that it's missing an IP:
```bash
# ip address show dev nodelocaldns
23: nodelocaldns: <BROADCAST,NOARP,UP,LOWER_UP> mtu 1300 qdisc noqueue state UNKNOWN group default 
    link/ether 72:b5:c3:81:f6:2f brd ff:ff:ff:ff:ff:ff
    inet6 fe80::70b5:c3ff:fe81:f62f/64 scope link 
       valid_lft forever preferred_lft forever
```

negz:
Scratch that - I didn't read the code closely enough and confused `EnsureDummyDevice`, which does not check for errors when setting an IP, with `AddDummyDevice` which does.

Nevertheless, somehow the device is being created without an IP address and also without producing any error while configuring the device.

dannyk81:
@negz hitting the same issue here (using CoreOS)

Were you able to resolve this?

dannyk81:
Something is very bizarre here... on one occasion, it was able to set the IP (just on one node, they are all identical and I didn't do any kind of changes):
```
$ kubectl logs -n kube-system nodelocaldns-mc52t
2019/01/19 17:04:34 2019-01-19T17:04:34.16Z [INFO] Tearing down
2019/01/19 17:04:34 2019-01-19T17:04:34.25Z [INFO] Setting up networking for node cache
ip6.arpa.:53 on 169.254.25.10
.:53 on 169.254.25.10
green.k8s.<domain>.com.:53 on 169.254.25.10
[...]

PR #286, opened 2019-01-23
Run checks before starting CoreDNS

prameshj:
https://github.com/kubernetes/dns/issues/282
There is sometimes a race in link creation and ip assignment.
If ip assignment is done too soon, the ip address does not persist.


dannyk81:
Thanks for this fix @prameshj! hope we can get this merged soon 😄

PR #291, opened 2019-03-09
Add retry logic when iptables checks fail

### LINKED ISSUE #292: node-local-dns crashes on failing to obtain iptables access
The upstream node-local-dns implementation (https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/dns/nodelocaldns) version `k8s.gcr.io/k8s-dns-node-cache:1.15.0` crashes frequently when used in a setup that has another application that also modifies iptables rules. 

I see this in my 1.10 cluster using weave-net cni that also runs on hostNetwork. Details of my setup are below:

```
$ weave version: 2.5.0 / 2.5.1

$ docker version
Client:
 Version:      17.03.2-ce
 API version:  1.27
 Go version:   go1.7.5
 Git commit:   f5ec1e2
 Built:        Tue Jun 27 02:09:56 2017
 OS/Arch:      linux/amd64

Server:
 Version:      17.03.2-ce
 API version:  1.27 (minimum version 1.12)
 Go version:   go1.7.5
 Git commit:   f5ec1e2
 Built:        Tue Jun 27 02:09:56 2017
 OS/Arch:      linux/amd64
 Experimental: false

$ uname -a
Linux ip-172-31-42-248 4.4.121-k8s #1 SMP Sun Mar 11 19:39:47 UTC 2018 x86_64 GNU/Linux

$ kubectl version
Server Version: version.Info{Major:"1", Minor:"10", GitVersion:"v1.10.3", GitCommit:"2bba0127d85d5a46ab4b778548be28623b32d0b0", GitTreeState:"clean", BuildDate:"2018-05-21T09:05:37Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"linux/amd64"}
```
With @prameshj 's help, we were able to see that when the nodelocaldns pod has issues with obtaining access to iptables it gives up too quickly with a locking error and then crashes:

```Failed to add back non-existent rule {raw PREROUTING [-p tcp -d 169.254.20.10 --dport 53 -j NOTRACK]} - error checking rule: exit status 4: Another app is currently holding the xtables lock. Stopped waiting after 2s.```

### PR DESCRIPTION
**Pull request by prameshj**
Nodelocaldns runs background checks to ensure its custom iptables rules are present. On failure, the pod restarts.
When deployed in an environment with other pods also using iptables heavily, there can be xtables lock contention. Adding a retry logic instead of exiting on first failure. 
Fixes #292

### DIFF
--- a/cmd/node-cache/main.go
+++ b/cmd/node-cache/main.go
@@ -7,6 +7,7 @@ import (
 	"net"
 	"os"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/coredns/coredns/coremain"
@@ -29,11 +30,12 @@ import (
 
 // configParams lists the configuration options that can be provided to dns-cache
 type configParams struct {
-	localIP   string        // ip address for the local cache agent to listen for dns requests
-	localPort string        // port to listen for dns requests
-	intfName  string        // Name of the interface to be created
-	interval  time.Duration // specifies how often to run iptables rules check
-	exitChan  chan bool     // Channel to terminate background goroutines
+	localIP            string        // ip address for the local cache agent to listen for dns requests
+	localPort          string        // port to listen for dns requests
+	intfName           string        // Name of the interface to be created
+	interval           time.Duration // specifies how often to run iptables rules check
+	maxIptablesLockErr int           // Max consecutive iptables lock errors to ignore before restarting
+	exitChan           chan bool     // Channel to terminate background goroutines
 }
 
 type iptablesRule struct {
@@ -50,6 +52,10 @@ type cacheApp struct {
 }
 
 var cache = cacheApp{params: configParams{localPort: "53"}}
+
+func IsIPTablesLocked(err error) bool {
+	return strings.Contains(err.Error(), "holding the xtables lock")
+}
 
 func (c *cacheApp) Init() {
 	err := c.parseAndValidateFlags()
@@ -156,6 +162,7 @@ func (c *cacheApp) parseAndValidateFlags() error {
 	flag.StringVar(&c.params.localIP, "localip", "", "ip address to bind dnscache to")
 	flag.StringVar(&c.params.intfName, "intfname", "nodelocaldns", "name of the interface to be created")
 	flag.DurationVar(&c.params.interval, "syncinterval", 60, "interval(in seconds) to check for iptables rules")
+	flag.IntVar(&c.params.maxIptablesLockErr, "maxiptableslockerr", 10, "Max consecutive iptables lock errors to ignore before restarting")
 	flag.Parse()
 
 	if net.ParseIP(c.params.localIP) == nil {
@@ -170,21 +177,30 @@ func (c *cacheApp) parseAndValidateFlags() error {
 	if _, err := strconv.Atoi(c.params.localPort); err != nil {
 		return fmt.Errorf("Invalid port specified - %q", c.params.localPort)
 	}
+	if c.params.maxIptablesLockErr < 0 {
+		return fmt.Errorf("Invalid value for maxIptablesLockErr - %d", c.params.maxIptablesLockErr)
+	}
 	return nil
 }
 
-func (c *cacheApp) runChecks() {
+// returns true if checks failed due to iptables lock issue
+func (c *cacheApp) runChecks() bool {
+	iptablesLocked := false
 	for _, rule := range c.iptablesRules {
 		exists, err := c.iptables.EnsureRule(utiliptables.Prepend, rule.table, rule.chain, rule.args...)
-		if !exists {
-			if err != nil {
-				cache.teardownNetworking()
-				clog.Fatalf("Failed to add back non-existent rule %v", rule)
-			}
+		if exists {
+			continue
+		} else if err == nil {
 			clog.Infof("Added back nonexistent rule - %v", rule)
+			continue
 		}
-		if err != nil {
-			clog.Errorf("Failed to check rule %v - %s", rule, err)
+		// if we got here, either iptables check failed or adding rule back failed.
+		if IsIPTablesLocked(err) {
+			clog.Infof("Failed to check/add back rule %v, due to xtables lock in use, retrying in %v seconds", rule, c.params.interval)
+			iptablesLocked = true
+		} else {
+			cache.teardownNetworking()
+			clog.Fatalf("Failed to add back non-existent rule %v - %s", rule, err)
 		}
 	}
 
@@ -199,15 +215,27 @@ func (c *cacheApp) runChecks() {
 	if err != nil {
 		clog.Errorf("Failed to check dummy device %s - %s", c.params.intfName, err)
 	}
+	return iptablesLocked
 }
 
 func (c *cacheApp) run() {
 	c.params.exitChan = make(chan bool, 1)
 	tick := time.NewTicker(c.params.interval * time.Second)
+	numLockErr := 0
+	iptablesLocked := false
 	for {
 		select {
 		case <-tick.C:
-			c.runChecks()
+			iptablesLocked = c.runChecks()
+			if iptablesLocked {
+				numLockErr++
+			} else {
+				numLockErr = 0
+			}
+			if numLockErr > c.params.maxIptablesLockErr {
+				cache.teardownNetworking()
+				clog.Fatalf("Too many consecutive failures when running iptables checks")
+			}
 		case <-c.params.exitChan:
 			clog.Warningf("Exiting iptables check goroutine")
 			return
--- a/cmd/node-cache/main.go
+++ b/cmd/node-cache/main.go
@@ -30,12 +30,12 @@ import (
 
 // configParams lists the configuration options that can be provided to dns-cache
 type configParams struct {
-	localIP            string        // ip address for the local cache agent to listen for dns requests
-	localPort          string        // port to listen for dns requests
-	intfName           string        // Name of the interface to be created
-	interval           time.Duration // specifies how often to run iptables rules check
-	maxIptablesLockErr int           // Max consecutive iptables lock errors to ignore before restarting
-	exitChan           chan bool     // Channel to terminate background goroutines
+	localIP     string        // ip address for the local cache agent to listen for dns requests
+	localPort   string        // port to listen for dns requests
+	metricsPort int           // port to serve node-cache metrics
+	intfName    string        // Name of the interface to be created
+	interval    time.Duration // specifies how often to run iptables rules check
+	exitChan    chan bool     // Channel to terminate background goroutines
 }
 
 type iptablesRule struct {
@@ -53,7 +53,7 @@ type cacheApp struct {
 
 var cache = cacheApp{params: configParams{localPort: "53"}}
 
-func IsIPTablesLocked(err error) bool {
+func isLockedErr(err error) bool {
 	return strings.Contains(err.Error(), "holding the xtables lock")
 }
 
@@ -75,6 +75,7 @@ func (c *cacheApp) Init() {
 		cache.teardownNetworking()
 		clog.Fatalf("Failed to setup - %s, Exiting", err)
 	}
+	initMetrics(net.JoinHostPort(c.params.localIP, strconv.Itoa(c.params.metricsPort)))
 }
 
 func init() {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: cmd/node-cache/main.go [hunk 1]
```diff
@@ -29,11 +30,12 @@ import (
...
-	localIP   string        // ip address for the local cache agent to listen for dns requests
-	localPort string        // port to listen for dns requests
-	intfName  string        // Name of the interface to be created
-	interval  time.Duration // specifies how often to run iptables rules check
-	exitChan  chan bool     // Channel to terminate background goroutines
+	localIP            string        // ip address for the local cache agent to listen for dns requests
+	localPort          string        // port to listen for dns requests
+	intfName           string        // Name of the interface to be created
+	interval           time.Duration // specifies how often to run iptables rules check
+	maxIptablesLockErr int           // Max consecutive iptables lock errors to ignore before restarting
```
bowei: when does it make sense to restart vs just keep try?

prameshj: I was thinking about this too. Since the lock error can occur when rules are being checked, it is possible that the rules are intact, but check failed, or the rules got removed and check failed as well. It would be good to recover at some point via restart - if rules cannot be added at startup, the pod will crashloop till it is added. It will cause dns downtime till the rules get added back - this is the case whether we restart or no.

bowei: Shouldn't we make the whole process convergent? I'm concerned about getting unlucky and ending up in the pod scheduling backoff.

prameshj: Made changes to update metrics counting setup errors rather than restarting.

cmd/node-cache/main.go
```diff
@@ -156,11 +163,13 @@ func (c *cacheApp) parseAndValidateFlags() error {
 	flag.StringVar(&c.params.localIP, "localip", "", "ip address to bind dnscache to")
 	flag.StringVar(&c.params.intfName, "intfname", "nodelocaldns", "name of the interface to be created")
 	flag.DurationVar(&c.params.interval, "syncinterval", 60, "interval(in seconds) to check for iptables rules")
+	flag.IntVar(&c.params.metricsPort, "metricsport", 9353, "port to serve nodecache setup metrics")
```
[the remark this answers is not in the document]
prameshj: 9353 port doesn't seem attached to a well-known service. We only create a listening socket on the nodelocalip:9353 address, so some other service should be able to use as long as binds to a specific ip:port.

cmd/node-cache/main.go
```diff
@@ -51,6 +53,10 @@ type cacheApp struct {
 
 var cache = cacheApp{params: configParams{localPort: "53"}}
 
+func IsIPTablesLocked(err error) bool {
```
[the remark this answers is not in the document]
prameshj: done

cmd/node-cache/main.go
```diff
@@ -176,28 +185,33 @@ func (c *cacheApp) parseAndValidateFlags() error {
 func (c *cacheApp) runChecks() {
 	for _, rule := range c.iptablesRules {
 		exists, err := c.iptables.EnsureRule(utiliptables.Prepend, rule.table, rule.chain, rule.args...)
-		if !exists {
-			if err != nil {
-				cache.teardownNetworking()
-				clog.Fatalf("Failed to add back non-existent rule %v", rule)
-			}
+		if exists {
+			continue
```
bowei: `clog.V(3).Infof("IP table rule (table %q, chain %q) already exists", ... )`

cmd/node-cache/main.go
```diff
@@ -29,11 +30,12 @@ import (
...
 type configParams struct {
-	localIP   string        // ip address for the local cache agent to listen for dns requests
-	localPort string        // port to listen for dns requests
-	intfName  string        // Name of the interface to be created
-	interval  time.Duration // specifies how often to run iptables rules check
-	exitChan  chan bool     // Channel to terminate background goroutines
+	localIP     string        // ip address for the local cache agent to listen for dns requests
+	localPort   string        // port to listen for dns requests
+	metricsPort int           // port to serve node-cache metrics
+	intfName    string        // Name of the interface to be created
```
bowei: interfaceName

cmd/node-cache/metrics.go
```diff
@@ -0,0 +1,46 @@
...
+	setupErrCount = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: plugin.Namespace,
+		Subsystem: "nodecache",
+		Name:      "setup_errors",
+		Help:      "The number of errors during periodic network setup for node-cache",
+	}, []string{"errortype"})
+)
+
+func initMetrics(ipport string) {
+	serveMetrics(ipport)
```
bowei: we should not eat the error here?

cmd/node-cache/metrics.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func registerMetrics() {
+	prometheus.MustRegister(setupErrCount)
+}
+
+func serveMetrics(ipport string) error {
+	ln, err := net.Listen("tcp", ipport)
+	if err != nil {
+		clog.Errorf("Failed to start metrics handler: %s", err)
```
bowei: if you are returning the error, then shouldn't log...

cmd/node-cache/main.go
```diff
@@ -176,28 +185,35 @@ func (c *cacheApp) parseAndValidateFlags() error {
...
 		}
 	}
 
 	exists, err := c.netifHandle.EnsureDummyDevice(c.params.intfName)
 	if !exists {
 		if err != nil {
-			cache.teardownNetworking()
-			clog.Fatalf("Failed to add back non-existent interface %s", c.params.intfName)
+			clog.Errorf("Failed to add back non-existent interface %s: %s", c.params.intfName, err)
+			setupErrCount.WithLabelValues("intf_add_err").Inc()
```
bowei: can we change the metric to interface_add_err?

cmd/node-cache/main.go
prameshj: Done

cmd/node-cache/metrics.go
prameshj: Fixed by checking error here and logging.

cmd/node-cache/metrics.go
prameshj: Fixed.

cmd/node-cache/main.go
prameshj: Fixed.

cmd/node-cache/main.go
```diff
@@ -176,28 +185,33 @@ func (c *cacheApp) parseAndValidateFlags() error {
 func (c *cacheApp) runChecks() {
 	for _, rule := range c.iptablesRules {
 		exists, err := c.iptables.EnsureRule(utiliptables.Prepend, rule.table, rule.chain, rule.args...)
-		if !exists {
-			if err != nil {
-				cache.teardownNetworking()
-				clog.Fatalf("Failed to add back non-existent rule %v", rule)
-			}
+		if exists {
```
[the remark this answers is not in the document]
prameshj: Done

cmd/node-cache/main.go
```diff
@@ -176,28 +185,33 @@ func (c *cacheApp) parseAndValidateFlags() error {
...
+			continue
+		} else if err == nil {
 			clog.Infof("Added back nonexistent rule - %v", rule)
+			continue
 		}
-		if err != nil {
-			clog.Errorf("Failed to check rule %v - %s", rule, err)
+		// if we got here, either iptables check failed or adding rule back failed.
+		if isLockedErr(err) {
+			clog.Infof("Failed to check/add back rule %v, due to xtables lock in use, retrying in %v seconds", rule, c.params.interval)
```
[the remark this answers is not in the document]
prameshj: Fixed

cmd/node-cache/main.go
```diff
@@ -176,28 +185,33 @@ func (c *cacheApp) parseAndValidateFlags() error {
...
 	for _, rule := range c.iptablesRules {
 		exists, err := c.iptables.EnsureRule(utiliptables.Prepend, rule.table, rule.chain, rule.args...)
-		if !exists {
-			if err != nil {
-				cache.teardownNetworking()
-				clog.Fatalf("Failed to add back non-existent rule %v", rule)
-			}
+		if exists {
+			continue
+		} else if err == nil {
```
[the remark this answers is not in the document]
prameshj: We need it because there are few additional cases. Fixed using a switch statement as suggested in a following comment.

cmd/node-cache/main.go
```diff
@@ -154,13 +161,15 @@ func (c *cacheApp) parseAndValidateFlags() error {
 	}
 
 	flag.StringVar(&c.params.localIP, "localip", "", "ip address to bind dnscache to")
-	flag.StringVar(&c.params.intfName, "intfname", "nodelocaldns", "name of the interface to be created")
+	flag.StringVar(&c.params.interfaceName, "intfname", "nodelocaldns", "name of the interface to be created")
```
bowei: fix the flag as well.

cmd/node-cache/main.go
```diff
@@ -176,28 +185,35 @@ func (c *cacheApp) parseAndValidateFlags() error {
...
-				cache.teardownNetworking()
-				clog.Fatalf("Failed to add back non-existent rule %v", rule)
-			}
+		switch {
+		case exists:
+			// debug messages can be printed by including "debug" plugin in coreFile.
+			clog.Debugf("IP table rule (table %q, chain %q) already exists", rule.table, rule.chain)
+			continue
+		case err == nil:
 			clog.Infof("Added back nonexistent rule - %v", rule)
```
bowei: Added nodelocaldns rule to table %q, chain %q

current log message is hard to understand unless you already know what is going on

cmd/node-cache/main.go
```diff
@@ -176,28 +185,35 @@ func (c *cacheApp) parseAndValidateFlags() error {
...
-		if err != nil {
-			clog.Errorf("Failed to check rule %v - %s", rule, err)
+			continue
+		// if we got here, either iptables check failed or adding rule back failed.
+		case isLockedErr(err):
+			clog.Infof("Failed to check/add back rule %v, due to xtables lock in use, retrying in %v", rule, c.params.interval)
+			setupErrCount.WithLabelValues("iptables_lock_error").Inc()
+		default:
+			clog.Errorf("Failed to add back non-existent rule %v - %s", rule, err)
+			setupErrCount.WithLabelValues("iptables_err").Inc()
```
bowei: be consistent with err vs error? change to iptables_error or just remove _error from all of the labels (the metric is already has error in its name, seems kind of redundant.

cmd/node-cache/main.go
```diff
@@ -176,28 +185,35 @@ func (c *cacheApp) parseAndValidateFlags() error {
...
-		}
-		if err != nil {
-			clog.Errorf("Failed to check rule %v - %s", rule, err)
+			continue
+		// if we got here, either iptables check failed or adding rule back failed.
+		case isLockedErr(err):
+			clog.Infof("Failed to check/add back rule %v, due to xtables lock in use, retrying in %v", rule, c.params.interval)
+			setupErrCount.WithLabelValues("iptables_lock_error").Inc()
+		default:
+			clog.Errorf("Failed to add back non-existent rule %v - %s", rule, err)
```
bowei: Error adding iptables rule (table %q, chain %q): %v

cmd/node-cache/main.go
```diff
@@ -176,28 +185,35 @@ func (c *cacheApp) parseAndValidateFlags() error {
...
 	}
 
-	exists, err := c.netifHandle.EnsureDummyDevice(c.params.intfName)
+	exists, err := c.netifHandle.EnsureDummyDevice(c.params.interfaceName)
 	if !exists {
 		if err != nil {
-			cache.teardownNetworking()
-			clog.Fatalf("Failed to add back non-existent interface %s", c.params.intfName)
+			clog.Errorf("Failed to add back non-existent interface %s: %s", c.params.interfaceName, err)
+			setupErrCount.WithLabelValues("interface_add_err").Inc()
```
bowei: interface_add_error or interface_add

cmd/node-cache/main.go
```diff
@@ -176,28 +185,35 @@ func (c *cacheApp) parseAndValidateFlags() error {
...
 		}
-		clog.Infof("Added back nonexistent interface - %s", c.params.intfName)
+		clog.Infof("Added back nonexistent interface - %s", c.params.interfaceName)
 	}
 	if err != nil {
-		clog.Errorf("Failed to check dummy device %s - %s", c.params.intfName, err)
+		clog.Errorf("Failed to check dummy device %s - %s", c.params.interfaceName, err)
+		setupErrCount.WithLabelValues("interface_check_err").Inc()
```
bowei: interface_check_error or interface_check

cmd/node-cache/metrics.go
```diff
@@ -0,0 +1,49 @@
...
+	setupErrCount = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: plugin.Namespace,
+		Subsystem: "nodecache",
+		Name:      "setup_errors",
+		Help:      "The number of errors during periodic network setup for node-cache",
+	}, []string{"errortype"})
+)
+
+func initMetrics(ipport string) {
+	err := serveMetrics(ipport)
```
bowei: combine?

if err := serveMetrics(...); err != nil {

cmd/node-cache/main.go
prameshj: It would be valuable to log the full rule so we know which exact rule (tcp/udp) failed.

cmd/node-cache/main.go
prameshj: Printing the rule which have the table/chain info as well. Modified the log message wording as suggested.

cmd/node-cache/main.go
bowei: yes, please add in the rule as well

cmd/node-cache/main.go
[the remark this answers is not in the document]
alvaroaleman: @prameshj How are ppl supposed to scrape this? That IP is only reachable from the node.

Are ppl supposed to run a proxy daemonset just to be able to gather metrics...? That sounds a bit impractical to me.

PR #294, opened 2019-03-28
Rebase container images from alpine to distroless.

yuwenma:
Updated containers:kube-dns, sidecar
Context:[KEP: Rebase k8s images to distroless](https://github.com/kubernetes/enhancements/pull/900)
Test: 
1. `make images` can create the following images
staging-k8s.gcr.io/k8s-dns-sidecar-amd64 
staging-k8s.gcr.io/k8s-dns-node-cache-amd64
staging-k8s.gcr.io/k8s-dns-kube-dns-amd64
staging-k8s.gcr.io/k8s-dns-dnsmasq-nanny-amd64
staging-k8s.gcr.io/k8s-dns-dnsmasq-amd64 

2. `docker run -e <with required flags like service host/port> <Kube-DNS-IMAGE-ID>`
can successfully upstart a container.

### DIFF
--- a/Dockerfile.node-cache
+++ b/Dockerfile.node-cache
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM debian:stable-slim
+FROM k8s.gcr.io/debian-base:v1.0.0
 RUN apt-get update && apt-get install -y \
     iproute2 \
     iptables \
--- a/rules.mk
+++ b/rules.mk
@@ -31,19 +31,19 @@ SRC_DIRS := cmd pkg
 ALL_ARCH := amd64 arm arm64 ppc64le s390x
 # Set default base image dynamically for each arch
 ifeq ($(ARCH),amd64)
-    BASEIMAGE?=alpine:3.8
+	BASEIMAGE?=k8s.gcr.io/debian-base:v1.0.0
 endif
 ifeq ($(ARCH),arm)
-    BASEIMAGE?=arm32v6/alpine:3.8
+	BASEIMAGE?=k8s.gcr.io/debian-base-arm:v1.0.0
 endif
 ifeq ($(ARCH),arm64)
-    BASEIMAGE?=arm64v8/alpine:3.8
+	BASEIMAGE?=k8s.gcr.io/debian-base-arm64:v1.0.0
 endif
 ifeq ($(ARCH),ppc64le)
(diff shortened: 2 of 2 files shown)


yuwenma:
Gentle ping. Can I get a review on this?

prameshj:
When switching to distroless, will we lose a shell on kube-dns containers?

yuwenma:
> When switching to distroless, will we lose a shell on kube-dns containers?

yes, we will.

MrHohn:
Tried building the new images and bring up a cluster with @prameshj, it seems like the sidecar is having issue coming up:
```
  Warning  Failed     2m (x4 over 3m)  kubelet, e2e-test-zihongz-minion-group-rq2k  Error: failed to start container "sidecar": Error response from daemon: linux spec user: unable to find user nobody: no matching entries in passwd file
```

yuwenma:
> no matching entries in passwd file

@Zihong, do you have any insights what this error may come from? I don't know what the  `nobody` USER does, but it seems to be specified in the sidecar [Dockerfile](https://github.com/kubernetes/dns/blob/eb273abe4447524312dca8742daa2bb343c54471/Dockerfile.sidecar#L21)

MrHohn:
@yuwenma I was under the impression that `nobody` was used just to make sure that container uses a user account that is in no privileged groups. Maybe that will not be a concern with distroless?

prameshj:
> > no matching entries in passwd file
> 
> @zihong, do you have any insights what this error may come from? I don't know what the `nobody` USER does, but it seems to be specified in the sidecar [Dockerfile](https://github.com/kubernetes/dns/blob/eb273abe4447524312dca8742daa2bb343c54471/Dockerfile.sidecar#L21)

The image works fine if the user is removed from the sidecar Dockerfile.

yuwenma:
According to the offline discussion, dns is currently rebased to kube custom debian-base instead of distroless for debugging convenience.

prameshj:
@yuwenma would you mind modifying Dockerfile.node-cache line 15 with the same debian image as well? We need to pick up fixes to CVEs, i was trying to see if we can change all base images in this same PR.

yuwenma:
> @yuwenma would you mind modifying Dockerfile.node-cache line 15 with the same debian image as well? We need to pick up fixes to CVEs, i was trying to see if we can change all base images in this same PR.

Done :smiley:

PR #299, opened 2019-04-28
Node cache: Make metrics listen address configurable

alvaroaleman:
This PR makes the metrics listen address, including interface, for the node cache configurable. This is required in order to actually be able to scrape those metrics, because the local IP is par design only reachable from the node the addon runs on.

The default of `127.0.0.2:9353` was chosen deliberately, as we can be very sure nothing will listen there already.

/assign @prameshj

### DIFF
--- a/cmd/node-cache/main.go
+++ b/cmd/node-cache/main.go
@@ -30,12 +30,12 @@ import (
 
 // configParams lists the configuration options that can be provided to dns-cache
 type configParams struct {
-	localIP       string        // ip address for the local cache agent to listen for dns requests
-	localPort     string        // port to listen for dns requests
-	metricsPort   int           // port to serve node-cache metrics
-	interfaceName string        // Name of the interface to be created
-	interval      time.Duration // specifies how often to run iptables rules check
-	exitChan      chan bool     // Channel to terminate background goroutines
+	localIP              string        // ip address for the local cache agent to listen for dns requests
+	localPort            string        // port to listen for dns requests
+	metricsListenAddress string        // address to serve metrics on
+	interfaceName        string        // Name of the interface to be created
+	interval             time.Duration // specifies how often to run iptables rules check
+	exitChan             chan bool     // Channel to terminate background goroutines
 }
 
 type iptablesRule struct {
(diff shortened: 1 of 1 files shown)


cmd/node-cache/main.go
```diff
@@ -163,7 +163,7 @@ func (c *cacheApp) parseAndValidateFlags() error {
 	flag.StringVar(&c.params.localIP, "localip", "", "ip address to bind dnscache to")
 	flag.StringVar(&c.params.interfaceName, "interfacename", "nodelocaldns", "name of the interface to be created")
 	flag.DurationVar(&c.params.interval, "syncinterval", 60, "interval(in seconds) to check for iptables rules")
-	flag.IntVar(&c.params.metricsPort, "metricsport", 9353, "port to serve nodecache setup metrics")
+	flag.StringVar(&c.params.metricsListenAddress, "metrics-listen-address", "127.0.0.2:9353", "address to serve metrics on")
```
prameshj: Was this value verified to work? Maybe we can make it listen on 0.0.0.0: by default?

alvaroaleman: Yes I did test it, it does work. The reason I didn't choose `0.0.0.0` by default is that there is a chance something is already listening on port 9353 which would make the cash crash.

LMK what you think.

prameshj: If some other service binds to 0.0.0.0:9353 first, this would still fail correct?

Instead, suppose a different service bound to <node-ip>:9353. And if this also exposed metrics on /metrics. What would accessing "http://node-ip:9353/metrics" give us? The metrics of node-local-dns or this other service?

Did you verify by using the node-ip to scrape the metrics?

alvaroaleman: > If some other service binds to 0.0.0.0:9353 first, this would still fail correct?

Yes, good point

> Instead, suppose a different service bound to :9353. And if this also exposed metrics on /metrics. What would accessing "http://node-ip:9353/metrics" give us? The metrics of node-local-dns or this other service?

The other service, obviously

alvaroaleman:
@prameshj I just realized that if the metrics port is already in use, the nodelocal dns cache will log an error but not fail to start. So I think it makes most sense to use `0.0.0.0:9353` as default and have changed it accordingly

PR #300, opened 2019-05-15
Upgrade CoreDNS to 1.5.0

woopstar:
In reference to https://github.com/coredns/coredns/issues/2636 there's a bug in CoreDNS under high load.

We have seen multiple `FORMERR` errors when talking to upstream DNS servers under version 1.2.6. The PR that fixes that in CoreDNS https://github.com/coredns/coredns/pull/2637 causes it to handle this problem correctly.

This is easy to replicate on a VMWare cluster where Promiscuous mode is enabled. When it's enabled, DNS resolution is very error-prone.


woopstar:
In reference to #274 I'll assign this to the same people:

/assign @justinsb
/assign @bowei
/assign @MrHohn
/assign @prameshj

chrisohaver:
> There are a few backward incompatible changes to plugins that are introduced between 1.2.6 and 1.5.0. Should ensure that none of these changes affect the current usage here (node local dns).

Reviewing release notes, the two potential impactful changes are...
1. `proxy` is removed as a plugin, use `forward` instead.
2. `health` now reports OK before the server is *ready*.  The `ready` plugin was added for the purpose of reporting readiness. IOW, liveness probes should use the `health` plugin, and readiness probes should use the `ready` plugin.

woopstar:
> > There are a few backward incompatible changes to plugins that are introduced between 1.2.6 and 1.5.0. Should ensure that none of these changes affect the current usage here (node local dns).
> 
> Reviewing release notes, the two potential impactful changes are...
> 
> 1. `proxy` is removed as a plugin, use `forward` instead.
> 2. `health` now reports OK before the server is _ready_.  The `ready` plugin was added for the purpose of reporting readiness. IOW, liveness probes should use the `health` plugin, and readiness probes should use the `ready` plugin.

I am unsure where to alter this. I agree though on the above!

bowei:
Could we split this PR into two commits: godep change, vendor/** autogenerated change.
Otherwise change looks ok to me.

/approve

bowei:
@prameshj can you do the final review/LGTM

prameshj:
> In reference to [coredns/coredns#2636](https://github.com/coredns/coredns/issues/2636) there's a bug in CoreDNS under high load.
> 
> We have seen multiple `FORMERR` errors when talking to upstream DNS servers under version 1.2.6. The PR that fixes that in CoreDNS [coredns/coredns#2637](https://github.com/coredns/coredns/pull/2637) causes it to handle this problem correctly.
> 
> This is easy to replicate on a VMWare cluster where Promiscuous mode is enabled. When it's enabled, DNS resolution is very error-prone.

Out of curiosity, was this problem observed with nodelocal dnscache as well? This uses TCP to talk to upstream and uses forward plugin which does reuse connection.. From the bug fix here, this looks like shared code for TCP and UDP - https://github.com/coredns/coredns/pull/2637/files, so in theory we could run into it. Curious to see if there were any tests that showed the problem with nodelocaldns.

prameshj:
> > There are a few backward incompatible changes to plugins that are introduced between 1.2.6 and 1.5.0. Should ensure that none of these changes affect the current usage here (node local dns).
> 
> Reviewing release notes, the two potential impactful changes are...
> 
> 1. `proxy` is removed as a plugin, use `forward` instead.
> 2. `health` now reports OK before the server is _ready_.  The `ready` plugin was added for the purpose of reporting readiness. IOW, liveness probes should use the `health` plugin, and readiness probes should use the `ready` plugin.

Thanks Chris! We already use the forward plugin and have liveness probes using health.
https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml#L134

Will keep this in mind when we add readiness probes.

PR #301, opened 2019-05-16
Upgrade dependencies to fix build

woopstar:
When #300 was introduced by me, the project was unable to build because of old dependencies.

This PR will add missing dependencies `dnstap/golang-dnstap` and `farsightsec/golang-framestream` which is required by CoreDNS 1.5.0. It will also update `client_golang` to version `0.9.2` which CoreDNS 1.5.0 uses.

The project is now able to build locally with `make build`


woopstar:
We have build the binaries and pushed a test image for linux to `pasientskyhosting/k8s-dns-node-cache-amd64:1.15.2-dirty`
We can confirm it works on our Kubernetes cluster where the node cache now starts and runs CoreDNS 1.5.0

prameshj:
Thanks for confirming this. Is Gopkg.toml file the only change made by hand?

woopstar:
Indeed. I only changed the toml file and added the two dependencies and upgraded the one. Nothing else was changed to make it build again.

PR #302, opened 2019-05-28
Add flag to skip iptables rules setup in nodecache

prameshj:
This can be used if iptables management is being offloaded to a different daemonset.
that manages iptables and other networking setup. HA using the approach described [here](https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/20190424-NodeLocalDNS-beta-proposal.md#design-details) becomes easier with this. All the iptables and liveness check logic resides in this other daemonset.

### DIFF
--- a/cmd/node-cache/main.go
+++ b/cmd/node-cache/main.go
@@ -45,6 +45,7 @@ type iptablesRule struct {
 }
 
 type cacheApp struct {
+	setupIptables bool
 	iptables      utiliptables.Interface
 	iptablesRules []iptablesRule
 	params        configParams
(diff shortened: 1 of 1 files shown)


carter-zhao:
@prameshj  Another daemonset component is needed  to dynamically determine when to use node-local-dns and when to flip to kube-dns endpoints [here](https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/20190424-NodeLocalDNS-beta-proposal.md).   I just can't find the code of the watcher component.

prameshj:
> @prameshj Another daemonset component is needed to dynamically determine when to use node-local-dns and when to flip to kube-dns endpoints [here](https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/20190424-NodeLocalDNS-beta-proposal.md). I just can't find the code of the watcher component.

This has been left as an implementer's choice right now - to identify which existing daemonset to pack this functionality into.  The logic will be along the lines of:

state = "enabled"
for {
    measure time taken for "dig @169.254.20.10 kubernetes.default.svc.cluster.local"
    if query times out 'n' consecutive times, remove NOTRACK iptables rules and set state to 'disabled'.
    if query successful 'n' successful times and current state is 'disabled', add the rules back.
    sleep 2s
}

sleep time and the threshold count can be customized.  This container would also manage setup of all iptables rules fornode-cache.


Are you looking for an image that can be run as a new daemonset?

carter-zhao:
@prameshj  Thank you for your reply

PR #303, opened 2019-05-29
Allow multiple listen ip addrs for node-cache

prameshj:
This is useful when implementing HA option as documented [here](https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/20190424-NodeLocalDNS-beta-proposal.md#design-details)

### DIFF
--- a/cmd/node-cache/main.go
+++ b/cmd/node-cache/main.go
@@ -30,7 +30,8 @@ import (
 
 // configParams lists the configuration options that can be provided to dns-cache
 type configParams struct {
-	localIP              string        // ip address for the local cache agent to listen for dns requests
+	localIPStr           string        // comma separated listen ips for the local cache agent
+	localIPs             []net.IP      // parsed ip addresses for the local cache agent to listen for dns requests
 	localPort            string        // port to listen for dns requests
 	metricsListenAddress string        // address to serve metrics on
 	interfaceName        string        // Name of the interface to be created
@@ -63,7 +64,7 @@ func (c *cacheApp) Init() {
 	if err != nil {
 		clog.Fatalf("Error parsing flags - %s, Exiting", err)
 	}
-	c.netifHandle = netif.NewNetifManager(net.ParseIP(c.params.localIP))
+	c.netifHandle = netif.NewNetifManager(c.params.localIPs)
 	if c.setupIptables {
 		c.initIptables()
 	}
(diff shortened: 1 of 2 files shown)


prameshj:
Verified that specifying multiple ips with the -localip flag works.

cmd/node-cache/main.go
```diff
@@ -167,15 +172,19 @@ func (c *cacheApp) parseAndValidateFlags() error {
...
+	flag.StringVar(&c.params.localIPStr, "localip", "", "comma-separated string of ip addresses to bind dnscache to")
 	flag.StringVar(&c.params.interfaceName, "interfacename", "nodelocaldns", "name of the interface to be created")
 	flag.DurationVar(&c.params.interval, "syncinterval", 60, "interval(in seconds) to check for iptables rules")
 	flag.StringVar(&c.params.metricsListenAddress, "metrics-listen-address", "0.0.0.0:9353", "address to serve metrics on")
 	flag.BoolVar(&c.setupIptables, "setupiptables", true, "indicates whether iptables rules should be setup")
 	flag.Parse()
 
-	if net.ParseIP(c.params.localIP) == nil {
-		return fmt.Errorf("Invalid localip specified - %q", c.params.localIP)
+	for _, ipstr := range strings.Split(c.params.localIPStr, ",") {
```
MrHohn: Should we error out when no IP is passed in (e.g. empty string)?

pkg/netif/netif.go
```diff
@@ -44,7 +50,13 @@ func (m *NetifManager) AddDummyDevice(name string) error {
 		return err
 	}
 	l, _ := m.LinkByName(name)
-	return m.AddrAdd(l, m.Addr)
+	for _, addr := range m.Addrs {
+		err = m.AddrAdd(l, addr)
+		if err != nil {
```
MrHohn: nit:
```golang
if err := m.AddrAdd(l, addr); err != nil {
    return err
}
```

pkg/netif/netif.go
```diff
@@ -23,7 +27,9 @@ func (m *NetifManager) EnsureDummyDevice(name string) (bool, error) {
 	l, err := m.LinkByName(name)
 	if err == nil {
 		// found dummy device, make sure ip matches. AddrAdd will return error if address exists, will add it otherwise
-		m.AddrAdd(l, m.Addr)
+		for _, addr := range m.Addrs {
+			m.AddrAdd(l, addr)
```
MrHohn: Check error here?

cmd/node-cache/main.go
prameshj: That case will also enter the loop since strings.Split returns the empty string. We will error out in line 185 below.

pkg/netif/netif.go
prameshj: AddrAdd will return error if the address already exists. That is a benign error, hence I did not include checking here. EnsureDummyDevice is called periodically. We do check the error in AddDummyDevice for the AddrAdd calls.

Issue #306, opened 2019-05-30
Upgrade to coreDNS 1.5 in node-cache


prameshj:
This is needed to pick up a bugfix coredns/coredns#2636

woopstar:
@prameshj can you test if our new image has the same problems with the cleanup that caused the revert:

`pasientskyhosting/k8s-dns-node-cache-amd64:1.15.4-2-g903339e-coredns1.6.2-dirty`

prameshj:
> @prameshj can you test if our new image has the same problems with the cleanup that caused the revert:
> 
> `pasientskyhosting/k8s-dns-node-cache-amd64:1.15.4-2-g903339e-coredns1.6.2-dirty`

It still does not do the cleanup.. This is how i test it:

1) Pick one of the node-local-dns pods and run "watch -n 1 'iptables -t raw -nvL'"
This command watches the iptables rules that node-local-dns pod configures.
2) Delete that specific node-local-dns pod with "kubectl delete pods <name>"
3) See the output of 1) if the rules are still there.

I see the rules are still there and get cleaned up when the next pod comes up.  This is problematic when we disable nodelocaldns and the rules/interface are not cleaned up.

Last I had checked, like the shutdown callbacks in forward plugin are taking longer/not completing before pod is shut down. The node-cache cleanup would be called after all the shutdown callbacks of plugins. @johnbelamaric any ideas on how to debug this further?

jordips:
Ok. I test it with the image and PREROUTING rules are not deleted, as you said 

```
Chain PREROUTING (policy ACCEPT 4239 packets, 952K bytes)
    4   374 CT         udp  --  *      *       0.0.0.0/0            169.254.25.10        udp dpt:53 NOTRACK
    0     0 CT         tcp  --  *      *       0.0.0.0/0            169.254.25.10        tcp dpt:53 NOTRACK
```
I will try to debug a litt bit more.

jordips:
> Last I had checked, like the shutdown callbacks in forward plugin are taking longer/not completing before pod is shutdown

@prameshj How did you debug this part? I'm trying to see logs when pod is deleted but I can't access with --previous flag. 

Thank you

prameshj:
I think i added logs to the different plugin shutdown callbacks and checked if they are invoked.
I started with logs here - https://github.com/kubernetes/dns/blob/64023fb045bd936c78800c98fceb3445113f51c1/vendor/github.com/mholt/caddy/sigtrap_posix.go#L45
[...]

PR #307, opened 2019-05-30
Revert upgrade to CoreDNS 1.5.0

prameshj:
The new image is not running the node-cache cleanup routine which deletes iptables rules and local interface created for node-cache. This is a problem when disabling the addon, if there is left over state.

Looks like the shutdown callbacks in forward plugin are taking longer/not completing before pod is shut down. The node-cache cleanup would be called after this. 
Reverting the upgrade to get the image into k/k before code freeze.
Created an issue to rootcause this and upgrade to 1.5.0 - https://github.com/kubernetes/dns/issues/306


axot:
Hello, is any updates for this?

prameshj:
We need atleast prometheus version 0.9.3, but the current skydns code has a constraint on version 0.8.0. I submitted https://github.com/skynetservices/skydns/pull/350 for review, but have not gotten an update. I will modify this repo to use a fork of skydns and modify it there.

axot:
Thank you. How about the bug around `OnProcessExit` callback, it seems not works after `1.5.0`?

prameshj:
> Thank you. How about the bug around `OnProcessExit` callback, it seems not works after `1.5.0`?

I haven't had a chance to debug that. Will do as part of my ongoing change. In the meantime, if you have any insights to share on the cleanup issue, that would be very helpful.

axot:
I did some tries, instead of appending to the tail of `caddy.OnProcessExit`, but insert into the head of the list. It did not work properly, so I'm going to create a simple project to easy debug this issue, at the same time, I found there are some cases that node-local-node totally not working. it may be related to `hostNetwork: true, dnsPolicy: ClusterFirstWithHostNet`, the condition is not clarified, I will fire an issue of this in the next week.

PR #322, opened 2019-09-05
Support kube-dns stubdomain config in node-cache

prameshj:
This change adds functionality in node-cache to generate the Corefile based on stubDomain and upstream servers config in kube-dns config map. This requires the kube-dns configmap to be mounted in the local filesystem.

The nodelocaldns yaml will look like this - https://github.com/kubernetes/kubernetes/commit/af3316b6c4d58814d86bb4b40ebfe80ddc949659

### DIFF
--- a/cmd/node-cache/app/cache_app.go
+++ b/cmd/node-cache/app/cache_app.go
@@ -73,6 +73,8 @@ func (c *CacheApp) Init() {
 		clog.Fatalf("Failed to setup - %s, Exiting", err)
 	}
 	initMetrics(c.params.MetricsListenAddress)
+	// Write the config file from template.
+	// this is required in case there is no kube-dns configmap specified.
 	c.updateCorefile(&config.Config{})
 	c.initKubeDNSConfigSync()
 }
--- a/cmd/node-cache/app/cache_app.go
+++ b/cmd/node-cache/app/cache_app.go
@@ -1,0 +1,242 @@
+package app
+
+import (
+	"fmt"
+	"os"
+
+	"k8s.io/dns/cmd/kube-dns/app/options"
+	"k8s.io/dns/pkg/dns/config"
+
+	"net"
+	"strings"
+	"time"
+
+	"github.com/coredns/coredns/coremain"
+	clog "github.com/coredns/coredns/plugin/pkg/log"
+	"k8s.io/dns/pkg/netif"
+	"k8s.io/kubernetes/pkg/util/dbus"
+	utilexec "k8s.io/kubernetes/pkg/util/exec"
+	utiliptables "k8s.io/kubernetes/pkg/util/iptables"
+)
+
+// ConfigParams lists the configuration options that can be provided to node-cache
+type ConfigParams struct {
+	LocalIPStr           string        // comma separated listen ips for the local cache agent
+	LocalIPs             []net.IP      // parsed ip addresses for the local cache agent to listen for dns requests
+	LocalPort            string        // port to listen for dns requests
+	MetricsListenAddress string        // address to serve metrics on
+	InterfaceName        string        // Name of the interface to be created
+	Interval             time.Duration // specifies how often to run iptables rules check
+	BaseCoreFile         string        // Path to the template config file for node-cache
+	CoreFile             string        // Path to config file used by node-cache
+	KubednsCMPath        string        // Directory where kube-dns configmap will be mounted
+	UpstreamSvcName      string        // Name of the service whose clusterIP is the upstream for node-cache for cluster domain
+	SetupIptables        bool
+}
+
+type iptablesRule struct {
+	table utiliptables.Table
+	chain utiliptables.Chain
+	args  []string
+}
+
+// CacheApp contains all the config required to run node-cache.
+type CacheApp struct {
+	iptables      utiliptables.Interface
+	iptablesRules []iptablesRule
+	params        *ConfigParams
+	netifHandle   *netif.NetifManager
+	kubednsConfig *options.KubeDNSConfig
+	exitChan      chan bool // Channel to terminate background goroutines
+	clusterDNSIP  net.IP
+}
+
+func isLockedErr(err error) bool {
+	return strings.Contains(err.Error(), "holding the xtables lock")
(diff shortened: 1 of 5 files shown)


prameshj:
@bowei as per offline discussion, I added a unit test to verify stubDomains support in node-cache. PTAL, thanks.

cmd/node-cache/app/configmap.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	baseConfig = bytes.Replace(baseConfig, []byte("__PILLAR__UPSTREAM__SERVERS__"), []byte(upstreamServers), -1)
+	baseConfig = bytes.Replace(baseConfig, []byte("__PILLAR__CLUSTER__DNS__"), []byte(c.clusterDNSIP.String()), -1)
+	baseConfig = bytes.Replace(baseConfig, []byte("__PILLAR__LOCAL__DNS__"), []byte(c.params.LocalIPStr), -1)
+
+	newConfig := bytes.Buffer{}
+	newConfig.WriteString(string(baseConfig))
+	newConfig.WriteString(stubDomainStr)
+	err = ioutil.WriteFile(c.params.CoreFile, newConfig.Bytes(), 0666)
+	if err != nil {
```
MrHohn: nit:
```golang
if err := ioutil.WriteFile(c.params.CoreFile, newConfig.Bytes(), 0666); err != nil {
...
```

cmd/node-cache/app/cache_app.go
```diff
@@ -0,0 +1,242 @@
...
+		// Logging error at info level
+		clog.Infof("Hit error during teardown - %s", err)
+	}
+	err = c.setupNetworking()
+	if err != nil {
+		c.TeardownNetworking()
+		clog.Fatalf("Failed to setup - %s, Exiting", err)
+	}
+	initMetrics(c.params.MetricsListenAddress)
+	c.updateCorefile(&config.Config{})
```
MrHohn: Curious why do we need to update it once with an empty config? It seems `initKubeDNSConfigSync` also updates it with the initialConfig.

MrHohn: Ah I see, that may not run if configmap path is not given.

cmd/node-cache/app/configmap.go
```diff
@@ -0,0 +1,103 @@
...
+		clog.Errorf("Failed to write config file %s - err %v", c.params.CoreFile, err)
+		return
+	}
+	clog.Infof("Updated Corefile with %d custom stubdomains and upstream servers %s", len(dnsConfig.StubDomains), upstreamServers)
```
MrHohn: Would it be helpful to log the actual upstreamServers/stubDomains instead of just the number?

MrHohn: (Was thinking may be useful when correlating events on customer's cluster.)

cmd/node-cache/app/cache_app_test.go
```diff
@@ -0,0 +1,162 @@
...
+		if err != nil {
+			t.Errorf("Failed to write stubDomains file - %s, err %v", upstreamServerFileName, err)
+		}
+	}
+	return ""
+}
+
+func createBaseFiles(t *testing.T, p *ConfigParams) {
+	err := ioutil.WriteFile(p.BaseCoreFile, []byte(templateCoreFileContents), os.ModePerm)
+	if err != nil {
```
MrHohn: nit:
```golang
if err := ioutil.WriteFile(p.BaseCoreFile, []byte(templateCoreFileContents), os.ModePerm); err != nil {
...
```

cmd/node-cache/app/cache_app_test.go
```diff
@@ -0,0 +1,162 @@
...
+		t.Fatalf("Failed to write template config file - %v", err)
+	}
+	if err = os.Mkdir(p.KubednsCMPath, os.ModePerm); err != nil {
```
MrHohn: nit: `if err := ...`

cmd/node-cache/app/cache_app_test.go
```diff
@@ -0,0 +1,162 @@
...
+	upstreamServerFileName = "upstreamNameservers"
+	UpstreamClusterDNS     = "test-svc"
+)
+
+func updateStubDomainsAndUpstreamServers(t *testing.T, p *ConfigParams, c *config.Config) string {
+	if stubDomainBlob, err := json.Marshal(c.StubDomains); err != nil {
+		t.Errorf("Failed to marshal stubdomains info, err %v", err)
+	} else {
+		err = ioutil.WriteFile(filepath.Join(p.KubednsCMPath, stubDomainFileName), stubDomainBlob, os.ModePerm)
+		if err != nil {
```
MrHohn: nit: `if err:= ...; err != nil {`

cmd/node-cache/app/cache_app_test.go
```diff
@@ -0,0 +1,162 @@
...
+			t.Errorf("Failed to write stubDomains file - %s, err %v", stubDomainFileName, err)
+		}
+	}
+
+	if upstreamBlob, err := json.Marshal(c.UpstreamNameservers); err != nil {
+		t.Errorf("Failed to marshal upstream nameservers info, err %v", err)
+	} else {
+		err = ioutil.WriteFile(filepath.Join(p.KubednsCMPath, upstreamServerFileName), upstreamBlob, os.ModePerm)
+		if err != nil {
```
MrHohn: nit: `if err:= ...; err != nil {`

cmd/node-cache/app/configmap.go
prameshj: It would be.. but i was wondering if it is worth logging if there are several hundred stubDomains configured.

cmd/node-cache/app/cache_app.go
prameshj: exactly. added a comment here to clarify this.

cmd/node-cache/app/configmap.go
prameshj: I chaned the code to log the updated coreFile at info level.

PR #327, opened 2019-10-10
don't hardcode coredns version

dwradcliffe:
The version hardcoded here is already out of date. This pulls the version string from the upstream package.

help


prameshj:
/lgtm
Updating the version is tracked in https://github.com/kubernetes/dns/issues/306

PR #325, opened 2019-10-08
Upgrade CoreDNS to v1.6.4

chad-jones:
As requested by @prameshj  in issue #306

This is using our fork of [skydns](https://github.com/pasientskyhosting/skydns) with updated dependencies.

PR #328, opened 2019-10-11
Upgrade to CoreDNS 1.6.4 in node-cache.

prameshj:
This fixes https://github.com/kubernetes/dns/issues/306

This PR uses a fork of skydns - https://github.com/pasientskyhosting/skydns with all the dependencies fixed.
Thanks @chad-jones, I used most of your changes from https://github.com/kubernetes/dns/pull/325, but the one change i did make is moving all constraints to Gopkg.toml, so there are no manual changes in Gopkg.lock

Verified that node-local-dns pods cleanup on exit.

Marking this PR on hold till we figure out if we can fix the original repo - https://github.com/skynetservices/skydns/pull/350 or create another fork.

/hold


prameshj:
Upstream skydns has been updated. Verified that the new node-local-dns pods are cleaning up rules upon exit.
/hold cancel

prameshj:
new image is at k8s.gcr.io/k8s-dns-node-cache:1.15.8

Issue #315, opened 2019-07-15
Add support for AAAA records in dns specification


sftim:
[DNS specification 1.0.1](https://github.com/kubernetes/dns/blob/cf7588d7932db067f376c5903e8b21fca769a4ba/docs/specification.md) covers IPv4 via A records, and does not mention IPv6 support via AAAA records.

As a feature request, add details to the specification that explain how to use DNS service discovery with IPv6.

bowei:
Would be great to link this from the IPv6 KEPs/todo list.

aojea:
I thought that KubeDNS was deprecated and replaced by CoreDNS, that has support for AAAA records since https://github.com/kubernetes/kubernetes/pull/44403

chrisohaver:
This issue refers to updating the k8s cluster dns spec itself, which does not currently mention AAAA records.

The update could be as simple as replacing `A` with `A/AAAA`, in the document.

bowei:
cc: @lachie83 
Is this tracked on the IPv6 work?

lachie83:
Investigating

aojea:
@bowei @lachie83 I can take this and submit a PR to modify the DNS specification for IPv6
/assign

lachie83:
Thanks @aojea

PR #335, opened 2019-12-15
Add IPv6 support to the DNS specification

aojea:
Add IPv6 support to the DNS specification:

* add AAAA records for IPv6 addresses
* add IP6.ARPA domain for IPv6 reverse lookup

It also removes the section `2.6 - Deprecated Records`
because the deprecated records are IPv4 only 

Ref: https://tools.ietf.org/html/rfc3596

Fixes: #315

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -11,7 +11,7 @@ is intended to provide a baseline for commonality between implementations.
 
 ## 1 - Schema Version
 
-This document describes version 1.0.1 of the schema.
+This document describes version 1.1.0 of the schema.
 
 ## 2 - Resource Records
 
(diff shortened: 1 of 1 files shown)


docs/specification.md
```diff
@@ -11,7 +11,7 @@ is intended to provide a baseline for commonality between implementations.
 
 ## 1 - Schema Version
 
-This document describes version 1.0.1 of the schema.
+This document describes version 1.0.2 of the schema.
```
aojea: What should be the right schema version?

docs/specification.md
```diff
@@ -184,14 +253,16 @@ exist.
   - `foo.default.svc.cluster.local. 10 IN CNAME www.example.com.`
   - `www.example.com. 28715 IN A 192.0.2.53`
 
-### 2.6 - Deprecated Records
-Kube-DNS versions prior to implementation of this specification also replied
-with an `A` record of the form below for any values of `<a>`, `<b>`, `<c>`, and `<d>` between 0 and 255:
```
aojea: I don't know if it's ok to remove the old records, but seems reasonable because the deprecation was announced in Dec 2016 based on git blame.
My rationale is that because they are IPv4 only it may be problematic having different records for each IP family.

docs/specification.md
thockin: I think 1.1 ?

docs/specification.md
bowei: We should send something out on the sig-network mailing list as FYI
Did the various implementations remove this record yet?

bowei:
/lgtm
cc: @prameshj 

@johnbelamaric has the final sign off

bowei:
Actually 

/approve

We can fix it up w/ a patch if there are comments.

sftim:
Thanks @aojea!

docs/specification.md
johnbelamaric: No, they all still support it. It's a config option in CoreDNS, I don't think there is any way to turn it off in kube-dns. I don't know about, say, OpenShift which I believe has its own implementation (at least it used to).

johnbelamaric:
For posterity sake
/approve

The only question is whether we change kube-dns to no longer provide the deprecated records. I think that's not really necessary, since they are additional records outside the spec, and that is allowed.

We could change the default CoreDNS configuration to not include those records for new clusters, but as @bowei mentioned it should be sent out to sig-network.

PR #340, opened 2020-01-17
Fixed negative Prometheus counter panic in DNS sidecar. 

sambdavidson:
Fixed source of crash looping of DNS in case of dnsmasq connection failures. 

While working on some networking issues within some clusters I noticed one of the dns pods was crash looping:

```stderr
...
127.0.0.1:58737->127.0.0.1:53: read: connection refused
W0115 22:13:33.013395       1 server.go:64] Error getting metrics from dnsmasq: read udp 127.0.0.1:41583->127.0.0.1:53: read: connection refused
W0115 22:13:38.013747       1 server.go:64] Error getting metrics from dnsmasq: read udp 127.0.0.1:56871->127.0.0.1:53: read: connection refused
W0115 22:13:43.014078       1 server.go:64] Error getting metrics from dnsmasq: read udp 127.0.0.1:37919->127.0.0.1:53: read: connection refused
W0115 22:13:48.018356       1 server.go:64] Error getting metrics from dnsmasq: read udp 127.0.0.1:48715->127.0.0.1:53: read: connection refused
W0115 22:16:03.039440       1 server.go:64] Error getting metrics from dnsmasq: read udp 127.0.0.1:54071->127.0.0.1:53: read: connection refused
panic: counter cannot decrease in value

goroutine 1 [running]:
k8s.io/dns/vendor/github.com/prometheus/client_golang/prometheus.(*counter).Add(0xc000427a00, 0xc05f400000000000)
    /go/src/k8s.io/dns/vendor/github.com/prometheus/client_golang/prometheus/counter.go:71 +0xa3
k8s.io/dns/pkg/sidecar.exportMetrics(0xc00000c088)
    /go/src/k8s.io/dns/pkg/sidecar/server.go:86 +0x1b0
k8s.io/dns/pkg/sidecar.(*server).runMetrics(0xc000019950, 0xc0001367e0)
    /go/src/k8s.io/dns/pkg/sidecar/server.go:68 +0x1b4
k8s.io/dns/pkg/sidecar.(*server).Run(0xc000019950, 0xc0001367e0)
    /go/src/k8s.io/dns/pkg/sidecar/server.go:53 +0x233
main.main()
    /go/src/k8s.io/dns/cmd/sidecar/main.go:56 +0x1dc
```

The error is caused because within [line 86 of pkg/sidecar/server.go]( https://github.com/kubernetes/dns/blame/master/pkg/sidecar/server.go#L86) we call `Counter.Add(...)` with a negative value because the assumption is made that every new value will be greater than the previous. It seems when there are connectivity issues this isn't the case.

When [the counter Add() function within Prometheus](https://github.com/prometheus/client_golang/blob/803ef2a759d7caaaa0de58e3815f1be4c8b5a42a/prometheus/counter.go#L89) is called with a negative value it panics. I believe this kills either the pod or at least some Prometheus goroutines. This a simple fix that checks for this case.

/assign mikedanese  bowei

### DIFF
--- a/pkg/sidecar/server.go
+++ b/pkg/sidecar/server.go
@@ -17,6 +17,7 @@ limitations under the License.
 package sidecar
 
 import (
+	"math"
 	"time"
 
 	"github.com/golang/glog"
@@ -79,10 +80,15 @@ func exportMetrics(metrics *dnsmasq.Metrics) {
 		// previous to get the proper value. This is needed because the
 		// Counter API does not allow us to set the counter to a value.
 		previousValue := countersCache[key]
-		delta := float64((*metrics)[key]) - previousValue
-		newValue := previousValue + delta
-		// Update cache to new value.
-		countersCache[key] = newValue
-		counters[key].Add(delta)
+		newValue := float64((*metrics)[key])
+		countersCache[key] = math.Max(newValue, 0)
+
+		// Ensure the newValue is a valid progression from the previous
+		// value. This will not be the case if for example the dnsmasq
+		// is experiencing connectivity issues. We can only call the
+		// counter Add(...) func with a positive delta between values.
+		if newValue > previousValue {
+			counters[key].Add(newValue - previousValue)
+		}
 	}
 }


prameshj:
I have a similar pending change here - https://github.com/kubernetes/dns/compare/master...prameshj:sidecar-check

The reason I haven't merged is because doing this check could disable metrics from being pushed for a long time. For instance, if dnsmasq had cache hits of say, 10,000 and then it restarted. sidecar did not restart, so it got a metric value of 0 and ignored it. Now it would ignore cache hits metric until it reaches 10k again. If that takes hours, then for all those hours, there will be no data on dnsmasq cache hits. Maybe the easier thing is for sidecar to panic and restart with fresh metrics?

sambdavidson:
@prameshj Oh okay interesting. Yeah, that is a good point. My opinion is to avoid a panic -> crash -> restart because I only found this bug when investigating other unrelated DNS issues. I think that restart loop would add a lot of unnecessary noise to the logs. A simple fix is to just reset "previous value" to 0 in the case of these errors. Then we don't have to wait for it to reset to 10k or whatever.

prameshj:
> @prameshj Oh okay interesting. Yeah, that is a good point. My opinion is to avoid a panic -> crash -> restart because I only found this bug when investigating other unrelated DNS issues. I think that restart loop would add a lot of unnecessary noise to the logs. A simple fix is to just reset "previous value" to 0 in the case of these errors. Then we don't have to wait for it to reset to 10k or whatever.

Yes, that makes sense. When I had looked into it, resetting the value involved invoking Run() on the sidecar again, I wasn't sure it was better than the sidecar restart. I agree that avoiding the pod panic/restart would help reduce noise in debugging. If you want to modify this PR handling the reset, I can take a look. Thanks!

sambdavidson:
@prameshj I have updated the code to always write the new value but conditionally add to the metric. PTAL.

pkg/sidecar/server.go
```diff
@@ -79,10 +80,15 @@ func exportMetrics(metrics *dnsmasq.Metrics) {
 		// previous to get the proper value. This is needed because the
 		// Counter API does not allow us to set the counter to a value.
 		previousValue := countersCache[key]
-		delta := float64((*metrics)[key]) - previousValue
-		newValue := previousValue + delta
-		// Update cache to new value.
-		countersCache[key] = newValue
-		counters[key].Add(delta)
+		newValue := float64((*metrics)[key])
+		countersCache[key] = math.Max(newValue, 0)
```
prameshj: If the previous value were 200 and new value is 250, would this set the value to 250 and in line 91, make the value 250 + (250 - 200) = 300?
That's not what we want, correct? 
Also, could you run this and trigger a dnsmasq restart and make sure sidecar exports the correct metrics?

sambdavidson: No I don't think so, I believe this is functionally identical. Previously with the two variables `delta` and `newValue` there was an odd bit of math:

```go
delta := float64((*metrics)[key]) - previousValue
newValue := previousValue + delta
```
Which is equal to:
```go
newValue := previousValue + (float64((*metrics)[key]) - previousValue)
```
The `previousValue` variables cancel out. So it is just a redundant bit of math I got rid of. 

Also regarding _running this and triggering a restart_, that is well outside of my experience as a Kubernetes (and linux dnsmasq) dev. I would need some guidance on how to do that. Something like:

1. rebuild the container
2. run it with docker or something?
3. ssh into the container and exec some magic command that forces a dnsmasq restart?

prameshj: Yes those are the steps. 
1) rebuild container and push to a dev registry with `make push REGISTRY=<name>`
2) I usually run this on an existing kubernetes cluster, i just change the image to the one i built. then ssh into dnsmasq and send SIGTERM via kill. 

Regarding the counter values, that explanation makes sense. So in case of reset, we will continue to increment dnsmasq stats from its previous value from before the crash. Would be good to verify it actually happens.

sambdavidson: Okay I will do those steps and attempt to test it next week on probably Tuesday. Thanks and hang tight.

sambdavidson: Hey @prameshj, sorry for being a month late with this. I have done the above steps and repeated the steps which previously caused crash looping in the DNS sidecar. There are now 0 pod restarts and the panics are gone from the logs. I think this is safe to submit. :)

prameshj: Thanks for checking. With this change, metrics will be reported when the new value is greater than the old value. If a dnsmasq restart happens and the value resets, no new metrics will be reported, until they catch up to the older value before dnsmasq restarted. Correct?

sambdavidson: No, previous value is always updated to `math.Max(newValue, 0)` no matter if it is greater than the previous value.

So if for instance the state is:

```go
countersCache[key] == 5 // previousValue will be set to 5 in line 82
```

If for the next two values for this key are `1` and `2` then the following will happen. For the call when `(*metrics)[key]` is equal to `1`:

```go
previousValue := countersCache[key] // 5
newValue := float64((*metrics)[key]) // 1
countersCache[key] = math.Max(newValue, 0) // updates countersCache[key] to 1
if newValue > previousValue { // 1 is not greater than 5 so we don't enter
  // counter add
}
```

Next for the call when `(*metrics)[key]` is equal to `2`:

```go
previousValue := countersCache[key] // 1
newValue := float64((*metrics)[key]) // 2
countersCache[key] = math.Max(newValue, 0) // updates countersCache[key] to 2
if newValue > previousValue { // 2 is greater than 1 so we enter
  // Counter .Add(...) with delta of (2-1)
}
```

So for that first instance where we have a huge decrease in value we don't log the negative delta, but in all subsequent calls we will take that new lower value as the next baseline.

prameshj: yes, makes sense. I missed that you update previous value every time.

Issue #321, opened 2019-08-21
k8s-dns-sidecar image specifies incorrect group


rtheis:
The k8s-dns-sidecar image specifies an incorrect group causing docker run to fail as is detailed below.  It appears that `USER nobody:nobody`, should be `USER nobody:nogroup` instead.

```
$ sudo docker run --rm -it --entrypoint=/bin/sh k8s.gcr.io/k8s-dns-sidecar-amd64:1.14.13
~ $ exit
$ sudo docker run --rm -it --entrypoint=/bin/sh k8s.gcr.io/k8s-dns-sidecar-amd64:1.15.4
docker: Error response from daemon: linux spec user: unable to find group nobody: no matching entries in group file.
$ sudo docker run --user nobody --rm -it --entrypoint=/bin/sh k8s.gcr.io/k8s-dns-sidecar-amd64:1.15.4
$ exit
$ sudo docker run --user nobody:nobody --rm -it --entrypoint=/bin/sh k8s.gcr.io/k8s-dns-sidecar-amd64:1.15.4
docker: Error response from daemon: linux spec user: unable to find group nobody: no matching entries in group file.
$ sudo docker run --user nobody:nogroup --rm -it --entrypoint=/bin/sh k8s.gcr.io/k8s-dns-sidecar-amd64:1.15.4
$ exit
```

PR #342, opened 2020-02-10
Debian base uses `nogroup` as the group

bowei:
Fixes: https://github.com/kubernetes/dns/issues/321

/assign @MrHohn @prameshj


prameshj:
Thanks for fixing this, verified that `docker run --rm -it gcr.io/pavithrar-k8s-dev/k8s-dns-sidecar-amd64:1.15.8-dirty` works.

PR #350, opened 2020-02-27
Use docker ARGs rather than sed

pnovotnak:
Simplify the build system using Docker ARG instructions instead of `sed`. Best reviewed commit-by-commit.

It’s currently a little bit of a pain to get the build working without make… Reddit uses drone, and using make + docker-in-docker is pretty clunky with that setup. This PR makes it a bit easier to use the dockerfiles without make

### DIFF
--- a/Dockerfile.dnsmasq-nanny
+++ b/Dockerfile.dnsmasq-nanny
@@ -11,10 +11,19 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-FROM ARG_REGISTRY/k8s-dns-dnsmasq-ARG_ARCH:ARG_VERSION
+
+ARG registry=staging-k8s.gcr.io
+ARG arch=amd64
+ARG version
+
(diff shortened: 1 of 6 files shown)


pnovotnak:
Sorry for the mess of force pushes

prameshj:
Could you rebase and fix the conflicts. I can take a look in the next couple of days. Sorry for the delay, I am not super familiar with these changes and  this somehow fell off my queue.

Dockerfile.dnsmasq-nanny
```diff
@@ -11,10 +11,19 @@
...
-FROM ARG_REGISTRY/k8s-dns-dnsmasq-ARG_ARCH:ARG_VERSION
+
+ARG registry=staging-k8s.gcr.io
+ARG arch=amd64
+ARG version
+
+FROM ${registry}/k8s-dns-dnsmasq-${arch}:${version}
+
+# Needs to be redefined in order to make this variable present during build
+ARG arch=amd64
```
prameshj, line 22: How would this now build for other ARCH that amd64? Previously we were passing in ARG_ARCH for each defined ARCH from rules.mk.

PR #353, opened 2020-03-03
Reuse the setup networking function in node-cache.

prameshj:
Use the same setup call for the initial setup as well as the periodic checks in node-cache.

### DIFF
--- a/cmd/node-cache/app/cache_app.go
+++ b/cmd/node-cache/app/cache_app.go
@@ -62,22 +62,18 @@ func (c *CacheApp) Init() {
 	if c.params.SetupIptables {
 		c.initIptables()
 	}
+	initMetrics(c.params.MetricsListenAddress)
+	// Write the config file from template.
+	// this is required in case there is no kube-dns configmap specified.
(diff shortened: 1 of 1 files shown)


prameshj:
@dannyk81 
/assign @MrHohn

cmd/node-cache/app/cache_app.go
```diff
@@ -126,26 +122,6 @@ func newIPTables() utiliptables.Interface {
 	return utiliptables.New(execer, dbus, utiliptables.ProtocolIpv4)
 }
 
-func (c *CacheApp) setupNetworking() error {
-	var err error
-	clog.Infof("Setting up networking for node cache")
-	err = c.netifHandle.AddDummyDevice(c.params.InterfaceName)
```
MrHohn: Just to make sure, so I assume the `EnsureDummyDevice()` func in the periodic check would have the same effect as this, hence not needing it any more?

prameshj: Yes, Ensure will call AddDummyDevice if it is not already present.

PR #361, opened 2020-03-26
Fix skipteardown flag not completely avoiding teardown

### LINKED ISSUE #358: skipteardown not working
Hi.

I'm using k8s.gcr.io/k8s-dns-node-cache:1.15.10 and the `-skipteardown` is not working.

```
2020/03/15 17:25:53 [INFO] Tearing down
2020/03/15 17:25:53 [INFO] Added back nodelocaldns rule - {raw PREROUTING [-p tcp -d 169.254.20.10 --dport 53 -j NOTRACK]}
2020/03/15 17:25:53 [INFO] Added back nodelocaldns rule - {raw PREROUTING [-p udp -d 169.254.20.10 --dport 53 -j NOTRACK]}
2020/03/15 17:25:53 [INFO] Added back nodelocaldns rule - {filter INPUT [-p tcp -d 169.254.20.10 --dport 53 -j ACCEPT]}
2020/03/15 17:25:53 [INFO] Added back nodelocaldns rule - {filter INPUT [-p udp -d 169.254.20.10 --dport 53 -j ACCEPT]}
2020/03/15 17:25:53 [INFO] Added back nodelocaldns rule - {raw OUTPUT [-p tcp -s 169.254.20.10 --sport 53 -j NOTRACK]}
2020/03/15 17:25:53 [INFO] Added back nodelocaldns rule - {raw OUTPUT [-p udp -s 169.254.20.10 --sport 53 -j NOTRACK]}
2020/03/15 17:25:53 [INFO] Added back nodelocaldns rule - {filter OUTPUT [-p tcp -s 169.254.20.10 --sport 53 -j ACCEPT]}
2020/03/15 17:25:53 [INFO] Added back nodelocaldns rule - {filter OUTPUT [-p udp -s 169.254.20.10 --sport 53 -j ACCEPT]}
2020/03/15 17:25:53 [INFO] Added back nodelocaldns rule - {raw OUTPUT [-p tcp -d 169.254.20.10 --dport 53 -j NOTRACK]}
2020/03/15 17:25:53 [INFO] Added back nodelocaldns rule - {raw OUTPUT [-p udp -d 169.254.20.10 --dport 53 -j NOTRACK]}
2020/03/15 17:25:53 [INFO] Added back nodelocaldns rule - {raw OUTPUT [-p tcp -d 169.254.20.10 --dport 8080 -j NOTRACK]}
2020/03/15 17:25:53 [INFO] Added back nodelocaldns rule - {raw OUTPUT [-p tcp -s 169.254.20.10 --sport 8080 -j NOTRACK]}
2020/03/15 17:25:53 [INFO] Added interface - nodelocaldns
cluster.local.:53 on 169.254.20.10
in-addr.arpa.:53 on 169.254.20.10
ip6.arpa.:53 on 169.254.20.10
.:53 on 169.254.20.10
[INFO] plugin/reload: Running configuration MD5 = 78a4f00fe5408e2d905f9c308b643b42
CoreDNS-1.6.7
linux/amd64, go1.11.13,
```

This  problem makes  inviable the HA setup because  the rollout of one DS will cause a restart on  the other when removing the interface.

For now, the team and I solve this issue putting two sets of DS on different pairs of IP/interfaces and using two nameservers on `/etc/resolv.conf` with the `rotate` option to ensure a better HA.

BTW, we are not using the kubernetes add-on but instead we implement this setup in conjunction with dnsPolicy and dnsConfig of the Pod Spec.

tommasopozzetti:
When using an HA setup of nodecache, it is imperative to avoid that one
instance of nodecache tear down networking at any time, to avoid disrupting
functionalities for the second instance, which is sharing the same interface
and iptables rules.

This fix addresses the issue of nodecache trying to tear down networking
at startup, right before ensuring it is setup again. When `--skipteardown`
flag is passed, nodecache should not try to do so, but instead simply
ensure that networking is correctly setup.

This should fix #358

### DIFF
--- a/cmd/node-cache/app/cache_app.go
+++ b/cmd/node-cache/app/cache_app.go
@@ -68,14 +68,6 @@ func (c *CacheApp) Init() {
 	// this is required in case there is no kube-dns configmap specified.
 	c.updateCorefile(&config.Config{})
 	c.initKubeDNSConfigSync()
-	if !c.params.SkipTeardown {
-		err := c.TeardownNetworking()
-		if err != nil {
-			// It is likely to hit errors here if previous shutdown cleaned up all iptables rules and interface.
-			// Logging error at info level
-			clog.Infof("Hit error during teardown - %s", err)
-		}
-	}
 	c.setupNetworking()
 }
 
--- a/cmd/node-cache/app/cache_app.go
+++ b/cmd/node-cache/app/cache_app.go
@@ -68,11 +68,13 @@ func (c *CacheApp) Init() {
 	// this is required in case there is no kube-dns configmap specified.
 	c.updateCorefile(&config.Config{})
 	c.initKubeDNSConfigSync()
-	err := c.TeardownNetworking()
-	if err != nil {
-		// It is likely to hit errors here if previous shutdown cleaned up all iptables rules and interface.
-		// Logging error at info level
-		clog.Infof("Hit error during teardown - %s", err)
+	if !c.params.SkipTeardown {
+		err := c.TeardownNetworking()
+		if err != nil {
+			// It is likely to hit errors here if previous shutdown cleaned up all iptables rules and interface.
+			// Logging error at info level
+			clog.Infof("Hit error during teardown - %s", err)
+		}
 	}
 	c.setupNetworking()
 }


cmd/node-cache/app/cache_app.go
```diff
@@ -68,11 +68,13 @@ func (c *CacheApp) Init() {
...
 	c.updateCorefile(&config.Config{})
 	c.initKubeDNSConfigSync()
-	err := c.TeardownNetworking()
-	if err != nil {
-		// It is likely to hit errors here if previous shutdown cleaned up all iptables rules and interface.
-		// Logging error at info level
-		clog.Infof("Hit error during teardown - %s", err)
+	if !c.params.SkipTeardown {
+		err := c.TeardownNetworking()
+		if err != nil {
```
MrHohn: nit:
```golang
if err := c.TeardownNetworking(); err != nil {
    ...
}
```

tommasopozzetti:
Hi @MrHohn,
thanks for your review.
I'm actually questioning, is the teardown even needed at all also for cases in which `--skipteardown` is not specified?
The line just following that one performs `setupNetworking()` which checks for the same rules and dummy interface that have just been destroyed by teardown and sets them up if not already existent. Even if a previous shutdown had not been clean, is it necessary to tear networking down to rebuild it right after?
Maybe that call to `TeardownNetworking()` could be deleted alltogether?

prameshj:
> Hi Zihong Zheng,
> thanks for your review.
> I'm actually questioning, is the teardown even needed at all also for cases in which `--skipteardown` is not specified?
> The line just following that one performs `setupNetworking()` which checks for the same rules and dummy interface that have just been destroyed by teardown and sets them up if not already existent. Even if a previous shutdown had not been clean, is it necessary to tear networking down to rebuild it right after?
> Maybe that call to `TeardownNetworking()` could be deleted alltogether?

It is necessary to keep it so the same iptables rules do not get appended over and over. Otherwise every restart will populate the chain with a duplicate rule.  Although I see we use "EnsureRule" which shouldn't add if it already exists. I haven't verified this recently, but I had seen these duplicate rules show up if teardown was skipped.

tommasopozzetti:
I pushed the change deleting the call alltogether. We can revert it back if something doesn't look right.
If avoiding teardown does leave duplicated iptables rules, however, that is something that should probably be addressed also in the case of `skipteardown` flag. Maybe rather than tearing down networking at startup, we could simply check for repeated rules and delete duplicates. That should work both with and without the `skipteardown` flag.

prameshj:
> I pushed the change deleting the call alltogether. We can revert it back if something doesn't look right.
> If avoiding teardown does leave duplicated iptables rules, however, that is something that should probably be addressed also in the case of `skipteardown` flag. Maybe rather than tearing down networking at startup, we could simply check for repeated rules and delete duplicates. That should work both with and without the `skipteardown` flag.

I did a quick test with these changes and by also commenting teardown upon exit - to see if duplicate rules get added. EnsureRule seems to be working correctly and adding rules only once.

PR #369, opened 2020-04-12
Add configmap error metric

povilasv:
Improve error handling a bit and add an error metric for config templating


prameshj:
Thanks!

/ok-to-test
/lgtm

PR #379, opened 2020-05-08
Declare all substitution strings as constants in node-cache

prameshj:
Also added a test yaml where the " __PILLAR__DNS__SERVER__" string is included without substitution. This should be replaced with empty string.
This is because __PILLAR__LOCAL__DNS__ includes both the listen IP for node-cache.

/assign @MrHohn

### DIFF
--- a/cmd/node-cache/app/configmap.go
+++ b/cmd/node-cache/app/configmap.go
@@ -31,6 +31,10 @@ const (
     forward . __PILLAR__UPSTREAM__SERVERS__
 `
 	DefaultConfigSyncPeriod = 10 * time.Second
+	UpstreamServerVar       = "__PILLAR__UPSTREAM__SERVERS__"
+	UpstreamClusterDNSVar   = "__PILLAR__CLUSTER__DNS__"
+	LocalListenIPsVar       = "__PILLAR__LOCAL__DNS__"
+	LocalDNSServerVar       = "__PILLAR__DNS__SERVER__"
 )
 
 // stubDomainInfo contains all the parameters needed to compute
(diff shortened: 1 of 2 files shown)


cmd/node-cache/app/configmap.go
```diff
@@ -77,20 +81,26 @@ func (c *CacheApp) updateCorefile(dnsConfig *config.Config) {
...
 		}
 	}
-	baseConfig = bytes.Replace(baseConfig, []byte("__PILLAR__CLUSTER__DNS__"), []byte(c.clusterDNSIP.String()), -1)
-	baseConfig = bytes.Replace(baseConfig, []byte("__PILLAR__LOCAL__DNS__"), []byte(strings.Replace(c.params.LocalIPStr, ",", " ", -1)), -1)
+	baseConfig = bytes.Replace(baseConfig, []byte(UpstreamClusterDNSVar), []byte(c.clusterDNSIP.String()), -1)
+	baseConfig = bytes.Replace(baseConfig, []byte(LocalListenIPsVar), []byte(strings.Replace(c.params.LocalIPStr, ",", " ", -1)), -1)
+	// All Listen IP Substitutions should have happened with replacing "LocalListenIPsVar". This is to ensure that no
+	// variables are left unsubstituted.
+	if bytes.Contains(baseConfig, []byte(LocalDNSServerVar)) {
+		baseConfig = bytes.Replace(baseConfig, []byte(LocalListenIPsVar), []byte(""), -1)
```
MrHohn, line 101: LocalListenIPsVar  -> LocalDNSServerVar?

prameshj: I called it ListenIPs because those are the IPs node-local-dns pods will listen for requests on.

Issue #389, opened 2020-07-02
Add NodeLocalDNS support for Azure Kubernetes (AKS)


luigibk:
Installation of NodeLocalDNS using the example manifest (minus the addon labels):

https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml

is successful on Azure Kubernetes (AKS) using the Azure CNI with kube-proxy in iptables mode. However dns resolution does not work anymore after the installation. 

At the end of the installation, on Azure (AKS running on Ubuntu 16.04 or Ubuntu 18.04) I have a nodelocaldns interface created:
```
nodelocaldns: <BROADCAST,NOARP,UP,LOWER_UP> mtu 1500 qdisc noqueue state UNKNOWN group default  
  link/ether 1a:a2:27:b6:2a:5e brd ff:ff:ff:ff:ff:ff
  inet 169.254.20.10/32 brd 169.254.20.10 scope global nodelocaldns
    valid_lft forever preferred_lft forever
  inet 10.0.0.10/32 brd 10.0.0.10 scope global nodelocaldns
    valid_lft forever preferred_lft forever  
``` 

where 10.0.0.10 is the cluster-dns. Also all the related iptables entries are created correctly:

```
*raw
:PREROUTING ACCEPT [40133:30257701]
:OUTPUT ACCEPT [5378:1385717]
-A PREROUTING -d 10.0.0.10/32 -p udp -m udp --dport 53 -j NOTRACK
-A PREROUTING -d 10.0.0.10/32 -p tcp -m tcp --dport 53 -j NOTRACK
-A PREROUTING -d 169.254.20.10/32 -p udp -m udp --dport 53 -j NOTRACK
-A PREROUTING -d 169.254.20.10/32 -p tcp -m tcp --dport 53 -j NOTRACK
-A OUTPUT -s 10.0.0.10/32 -p tcp -m tcp --sport 8080 -j NOTRACK
-A OUTPUT -d 10.0.0.10/32 -p tcp -m tcp --dport 8080 -j NOTRACK
-A OUTPUT -d 10.0.0.10/32 -p udp -m udp --dport 53 -j NOTRACK
-A OUTPUT -d 10.0.0.10/32 -p tcp -m tcp --dport 53 -j NOTRACK
-A OUTPUT -s 10.0.0.10/32 -p udp -m udp --sport 53 -j NOTRACK
[...]

PR #390, opened 2020-07-02
Nodecache ebtables option

luigibk:
Add extra `ebtables` option to NodeLocalDNS to address issue related to Azure Kubernetes (AKS):

https://github.com/kubernetes/dns/issues/389

This option, when true (default is false), adds the following 2 ebtables rules to nodes when NodeLocalDNS is installed:

```
sudo ebtables -t broute -A BROUTING --ip-dst 169.254.20.10 -p IPv4 -j redirect
sudo ebtables -t broute -A BROUTING --ip-dst 10.0.0.10 -p IPv4 -j redirect  
```

where the 2 IPs are from the -localip argument passed to NodeLocalDNS.

The 2 rules are automatically removed during NodeLocalDNS tear down as it happens already for iptables rules.

### DIFF
--- a/cmd/node-cache/app/cache_app.go
+++ b/cmd/node-cache/app/cache_app.go
@@ -17,6 +17,8 @@ import (
 	"k8s.io/kubernetes/pkg/util/dbus"
 	utilexec "k8s.io/kubernetes/pkg/util/exec"
 	utiliptables "k8s.io/kubernetes/pkg/util/iptables"
+	utilsexec "k8s.io/utils/exec"
+	utilebtables "k8s.io/utils/net/ebtables"
 )
 
 // ConfigParams lists the configuration options that can be provided to node-cache
@@ -33,6 +35,7 @@ type ConfigParams struct {
 	UpstreamSvcName      string        // Name of the service whose clusterIP is the upstream for node-cache for cluster domain
 	HealthPort           string        // port for the healthcheck
 	SetupIptables        bool
+	SetupEbtables        bool
 	SkipTeardown         bool // Indicates whether the iptables rules and interface should be torn down
 }
 
@@ -42,10 +45,18 @@ type iptablesRule struct {
 	args  []string
 }
 
+type ebtablesRule struct {
+	table utilebtables.Table
+	chain utilebtables.Chain
+	args  []string
+}
+
 // CacheApp contains all the config required to run node-cache.
 type CacheApp struct {
 	iptables      utiliptables.Interface
 	iptablesRules []iptablesRule
+	ebtables      utilebtables.Interface
+	ebtablesRules []ebtablesRule
 	params        *ConfigParams
 	netifHandle   *netif.NetifManager
 	kubednsConfig *options.KubeDNSConfig
@@ -62,6 +73,9 @@ func (c *CacheApp) Init() {
 	c.netifHandle = netif.NewNetifManager(c.params.LocalIPs)
 	if c.params.SetupIptables {
 		c.initIptables()
+	}
+	if c.params.SetupEbtables {
+		c.initEbtables()
 	}
 	initMetrics(c.params.MetricsListenAddress)
 	// Write the config file from template.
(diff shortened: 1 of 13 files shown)


prameshj:
Thanks for submitting this. This seems to be required for DNS requests to be redirected to NodeLocalDNS in AzureCNI implementation based on https://github.com/kubernetes/dns/issues/389

Can this not be included in the CNI plugin?

luigibk:
Hi @prameshj,

Thanks for getting back to me. I suppose it could be added to the CNI, even if I have not looked into the specific implementation details, so I cannot say for sure.
However my choice to add it in NodeLocalDNS has been driven by the following ideas:
- this seems quite specific to NodeLocalDNS and the CNI would use it just for that, as far as I can see.
- there could be other CNIs, implemented already or in future, that need this option for NodeLocalDNS to work.
- it's a fairly simple change, hidden behind an option, which in its default configuration does not change anything in the current NodeLocalDNS implementation.

luigibk:
Hi @prameshj, did you manage to have a look at this PR and think about my reply to your question? I am not sure I understand the idea that not all CNIs should be supported by NodeLocalDNS, especially when this support comes at the cost of a very small amount of code to maintain, introduced with a non-breaking change. It really raises the question of what makes a CNI more acceptable than others?

On this note, I don't know if you've had a look at the PR already, but the ebtables related part of it is already available in recent versions of kubernetes: https://github.com/kubernetes/kubernetes/pull/92440
which cannot be used in NodeLocalDNS because of some circular dependencies issue. I have opened a PR to move that part to kubernetes/utils: https://github.com/kubernetes/utils/pull/175 
which should solve that problem and reduce further the size of this PR by removing all the ebtables code. 

To be honest I think also the iptables dependency you have should not come from vendoring kubernetes, but moving that looks like a bigger job and for another conversation anyway.

prameshj:
> Hi @prameshj, did you manage to have a look at this PR and think about my reply to your question? I am not sure I understand the idea that not all CNIs should be supported by NodeLocalDNS, especially when this support comes at the cost of a very small amount of code to maintain, introduced with a non-breaking change. It really raises the question of what makes a CNI more acceptable than others?

My concern was adding all the ebtables utils to support working of NodeLocalDNS just in some setups. If the specific CNI is setting up ebtables for its working, adding in this logic when NodeLocalDNS is enabled, would likely be easier there.
Having said that, I am ok with including the flag change here, once the [kubernetes/utils#175](https://github.com/kubernetes/utils/pull/175) change merges and we vendor it here.

cc @bowei 
> 
> On this note, I don't know if you've had a look at the PR already, but the ebtables related part of it is already available in recent versions of kubernetes: [kubernetes/kubernetes#92440](https://github.com/kubernetes/kubernetes/pull/92440)
> which cannot be used in NodeLocalDNS because of some circular dependencies issue. I have opened a PR to move that part to kubernetes/utils: [kubernetes/utils#175](https://github.com/kubernetes/utils/pull/175)
> which should solve that problem and reduce further the size of this PR by removing all the ebtables code.
> 
> To be honest I think also the iptables dependency you have should not come from vendoring kubernetes, but moving that looks like a bigger job and for another conversation anyway.

luigibk:
@prameshj I've committed the changes agreed as https://github.com/kubernetes/utils/pull/175 has been merged.

cmd/node-cache/app/cache_app.go
```diff
@@ -170,6 +212,25 @@ func (c *CacheApp) setupNetworking() {
...
+	if c.params.SetupEbtables {
+		for _, rule := range c.ebtablesRules {
+			exists, err := c.ebtables.EnsureRule(utilebtables.Append, rule.table, rule.chain, rule.args...)
+			switch {
+			case exists:
+				// debug messages can be printed by including "debug" plugin in coreFile.
+				clog.Debugf("ebtables rule %v for nodelocaldns already exists", rule)
+				continue
+			case err == nil:
+				clog.Infof("Added back nodelocaldns rule - %v", rule)
```
prameshj, line 224: replace with "ebtables rule"

bowei:
/approve

Can you clean up the commits (squash) and make sure that the commit contains 1) how this flag is to be used 2) description in the PR above.


something like 

```
Add ebtables option to NodeLocalDNS

Add extra ebtables option to NodeLocalDNS to address issue related to Azure Kubernetes (AKS):

<usage>

Fixes #389

This option, when true (default is false), adds the following 2 ebtables rules to nodes when NodeLocalDNS is installed:

sudo ebtables -t broute -A BROUTING --ip-dst 169.254.20.10 -p IPv4 -j redirect
sudo ebtables -t broute -A BROUTING --ip-dst 10.0.0.10 -p IPv4 -j redirect  
where the 2 IPs are from the -localip argument passed to NodeLocalDNS.

The 2 rules are automatically removed during NodeLocalDNS tear down as it happens already for iptables rules.
```

luigibk:
Thanks @bowei Should be all done.

bowei:
thanks for cleaning up the commit message

/lgtm
/approve

PR #395, opened 2020-08-11
Update configmap.go to decouple from saltstack pillar names 

jayunit100:
see kubernetes/pull/93836


jayunit100:
@prameshj  xref https://github.com/kubernetes/kubernetes/pull/93836

jayunit100:
assuming this will fail until the 93836 patch merges...

jayunit100:
if 93936 will work by duplicating these records as opposed to just deleting the old ones, im happy to do that.

jayunit100:
Should we close this or leave it open? At some point I assume removing that pillar stuff will need to happen so maybe I'll leave it open....

PR #401, opened 2020-09-08
Log warning if kube-dns-upstream is not found.

prameshj:
For deployments where node-cache is non listening on kube-dns IP, the kube-dns-upstream
service is not relevant. This change adds a warning message instead of a fatal error.


MrHohn (approved):
/lgtm

PR #405, opened 2020-09-25
Fix warning message in node-cache when replacing custom upstream server

prameshj:
https://github.com/kubernetes/kubernetes/pull/93471 replaces the force_tcp block. So, it is expected for upstream block to not have TCP anymore.

Replacing TCP upstream block to UDP can fail if the existing config was
already using UDP.  Fix warning message to capture this.

/assign @MrHohn


MrHohn (approved):
/lgtm
/approve

PR #407, opened 2020-10-01
Adds option to NOT set up interface when deployed in non-root ns.

Weil0ng:
This is to support cases where we need to deploy node-local-dns as non-host pods.


Weil0ng:
I've verified locally that the new image with this PR works unchanged for existing setups (where this option is not set), and it works for intended use case (with Cilium to be specific) when this option is set to true.

prameshj:
> I've verified locally that the new image with this PR works unchanged for existing setups (where this option is not set), and it works for intended use case (with Cilium to be specific) when this option is set to true.

Thanks for verifying.

/lgtm
/approve

PR #400, opened 2020-09-08
Migrate from dep to Go modules and coreDNS 1.7

prameshj:
Building on top of https://github.com/kubernetes/dns/pull/384, fewer dependency changes and also no coreDNS upgrade yet.
Marking as WIP to get unit tests passing. Locally the tests fail with:

```
/tmp/go-build402259259/b518/sidecar.test flag redefined: log_dir
panic: /tmp/go-build402259259/b518/sidecar.test flag redefined: log_dir

goroutine 1 [running]:
flag.(*FlagSet).Var(0xc0000481e0, 0x140ec20, 0x1cb4c90, 0x12855af, 0x7, 0x12afc94, 0x2f)
	/usr/local/go/src/flag/flag.go:848 +0x4ae
flag.(*FlagSet).StringVar(...)
	/usr/local/go/src/flag/flag.go:751
k8s.io/klog.InitFlags(0xc0000481e0)
	/go/pkg/mod/k8s.io/klog@v1.0.0/klog.go:420 +0x7f
k8s.io/component-base/logs.init.0()
	/go/pkg/mod/k8s.io/component-base@v0.0.0-20190620085130-185d68e6e6ea/logs/logs.go:35 +0x2d
```

I suspect this is due to the GOPATH setting using gvm.

PR #408, opened 2020-10-06
Revert "Migrate from dep to Go modules and coreDNS 1.7"

prameshj:
Reverts kubernetes/dns#400

Will submit PR after investigating ginkgo build error

/assign @MrHohn


MrHohn (approved):
/lgtm
/approve

PR #413, opened 2020-10-13
Upgrade to Coredns 1.7 and migrate to go modules

prameshj:


prameshj:
/hold cancel

Tracis is now passing.

PR #418, opened 2020-10-31
Add ipv6 support to node-local-dns

aojea:
Add IPv6 support to node-local-dns, using the LocalIP to identify the IP family.
It validates that all the local-ips passed as parameter belong to the same IP family.
In case of requiring dual-stack, the user can launch 2 different node-local-dns process, one for each ip family.

Fixes: https://github.com/kubernetes/kubernetes/issues/93568


cmd/node-cache/app/cache_app_test.go
```diff
@@ -186,3 +186,79 @@ func TestUpdateCoreFile(t *testing.T) {
...
+
+func TestUpdateIPv6CoreFile(t *testing.T) {
+	baseDir, err := ioutil.TempDir("", "dnstest")
+	if err != nil {
+		t.Fatalf("Failed to obtain temp directory for testing, err %v", err)
+	}
+	envName := strings.ToUpper(strings.Replace(UpstreamClusterDNS, "-", "_", -1)) + "_SERVICE_HOST"
+	os.Setenv(envName, "2001:db8::1")
+	defer func() { os.RemoveAll(baseDir) }()
+	c, err := NewCacheApp(&ConfigParams{LocalIPStr: "fe80:169:254::1,fd00:1:2:3::5",
```
prameshj, line 198: Is "fd00:1:2:3::5" the v6 kube-dns IP and "fe80:169:254::1" the 169.254.20.10 counterpart? We should document this IP for v6 clusters, in https://kubernetes.io/docs/tasks/administer-cluster/nodelocaldns/#configuration

cmd/node-cache/app/cache_app.go
```diff
@@ -130,21 +137,29 @@ func (c *CacheApp) initIptables() {
...
-	return utiliptables.New(execer, dbus, utiliptables.ProtocolIpv4)
+	protocol := utiliptables.ProtocolIpv4
+	if isIPv6 {
+		protocol = utiliptables.ProtocolIpv6
+	}
+	return utiliptables.New(execer, dbus, protocol)
 }
 
 func (c *CacheApp) initEbtables() {
+	protocol := "IPV4"
```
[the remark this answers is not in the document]
aojea: good question, this is how it was before, see the diff in L147, but this is something I learned today, it seems that this reads `/etc/ethertypes` looking for the protocol and
>  The use of this file is not case sensitive
https://linux.die.net/man/8/ebtables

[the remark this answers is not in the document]
aojea: ahh, my bad, previously was with lowercase ... seems I need to change my glasses :)

cmd/node-cache/app/cache_app_test.go
aojea: My reading is that these are configuration options, right? An user not necessarily have to use those values, or do they?
I can update the docs using IPv6 addresses, but want to clarify that in IPv4 the link-local addresses are reserved to 169.254.0.0/16. For IPv6 the reserved range is fe80::/10, however, unlike IPv4, IPv6 requires a link-local address on every network interface on which the IPv6 protocol is enabled,.

IIRC 10.0.0.10 is the service ClusterIP, if that is right that will depend on the ServiceSubnet configured

aojea: I will submit a follow up PR to update the doc to mention you can use IPv6 addresses and adding some examples

prameshj: > My reading is that these are configuration options, right? An user not necessarily have to use those values, or do they?
True, the user can use any value for the listen ip. We do provide one example value for V4 - 169.254.20.10. It would be good to list one such example value for V6 and include a note that any linklocal IP is ok.

cmd/node-cache/app/cache_app.go
```diff
@@ -93,6 +94,12 @@ func (c *CacheApp) Init() {
 	c.params.SetupIptables = setupIptables
 }
 
+// isIPv6 return if the node-cache is working in IPv6 mode
+// LocalIPs are guaranteed to have the same family
+func (c *CacheApp) isIPv6() bool {
+	return utilnet.IsIPv6(c.params.LocalIPs[0])
```
prameshj, line 100: Can we handle the case where no localIPs are specified and return false in that case.

prameshj:
The change looks good to me. Was this verified to work in V6 as well as the default v4 case?

cmd/node-cache/app/cache_app.go
aojea: sure

cmd/node-cache/app/cache_app_test.go
aojea: https://github.com/kubernetes/website/pull/24957

aojea:
> The change looks good to me. Was this verified to work in V6 as well as the default v4 case?

do you run it in some CI? I can add a new job for IPv6 with KIND if you point me to a similar one
Otherwise I'll test it locally

sftim:
BTW, I commented in https://github.com/kubernetes/website/pull/24957#pullrequestreview-526709952 that the documentation for this didn't seem to make sense - the design doesn't seem right.

Sorry about that but it is a concern I have.

cmd/node-cache/app/cache_app_test.go
sftim: > IPv6 requires a link-local address on every network interface on which the IPv6 protocol is enabled

I'm trying to imagine what netns the node-local DNS could run in to make this work. Link-local traffic doesn't route, so does the node-local DNS run in the netns of the Pod? If so, which one? Maybe the node-local DNS runs in its own netns and something does route the link-local traffic to it?

It's hard to imagine - maybe there is a diagram that helps?

sftim: BTW I hope that 169.254.20.10 is a _documentation example_ and not a default. Unless you _know_ you control the link and the order that things arrive onto it, it's wrong to assume that you can pick any IPv4 address inside `169.254.0.0/16` and have that available. Given the diverse set of container network plugins, I don't think that assumption is safe.

sftim:
Sorry about not following this up yet

cmd/node-cache/app/cache_app_test.go
aojea: it' s a doc example, you can use whatever IP, it seems that the pod creates an interface with that IP inside the node, but there is nothing special about it

aojea:
@prameshj @sftim this is working, PTAL
```
cluster.local.:53 on fd00:8:8:8:8::8
cluster.local.:53 on fd00:10:96::1
in-addr.arpa.:53 on fd00:10:96::1
ip6.arpa.:53 on fd00:10:96::1
.:53 on fd00:10:96::1
[INFO] plugin/reload: Running configuration MD5 = 7019f851c438f92ca22fa86bd3960ac4
CoreDNS-1.7.0
linux/amd64, go1.13.15, 
```

sftim:
Sure, I'm happy for this PR to proceed.

cmd/node-cache/app/cache_app_test.go
prameshj: yes, it can be any IP as long as it is guaranteed to not collide with anything else on the node. Nothing else special about it.

prameshj (approved):

pkg/dns/config/config.go
```diff
@@ -95,15 +95,24 @@ func (config *Config) validateStubDomains() error {
 		}
 
 		for _, ns := range nsList {
-			// TODO(rramkumar): Use net.SplitHostPort to support ipv6 case.
-			nsStrings := strings.SplitN(ns, ":", 2)
+			var host, port string
+			var err error
+
+			host, port, err = net.SplitHostPort(ns)
```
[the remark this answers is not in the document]
aojea: yeah, ... now I wonder why I did that :thinking:

cmd/node-cache/main.go
```diff
@@ -83,6 +84,12 @@ func parseAndValidateFlags() (*app.ConfigParams, error) {
 		params.LocalIPs = append(params.LocalIPs, newIP)
 	}
 
+	// validate all the IPs have the same IP family
+	for _, ip := range params.LocalIPs {
+		if utilnet.IsIPv6(params.LocalIPs[0]) != utilnet.IsIPv6(ip) {
+			return params, fmt.Errorf("Invalid IP Family localip specified - %q", ip)
```
prameshj, line 90: Maybe reword to "Unexpected IP Family for localIP - %q, want IPv6=%v", ip, utilnet.IsIPv6(params.LocalIPs[0])

prameshj:
Verified that travis build passed - https://travis-ci.org/github/kubernetes/dns/builds/745675874

Will LGTM after the last comment is addressed. Thanks for adding v6 support, @aojea !

cmd/node-cache/main.go
aojea: yep, much better

aojea:
> Verified that travis build passed - https://travis-ci.org/github/kubernetes/dns/builds/745675874
> 
> Will LGTM after the last comment is addressed. Thanks for adding v6 support, @aojea !

you are welcome :)

aojea:
@prameshj I resubmitted the PR to retrigger the travis-ci job but is still not reporting and I've lost the lgtm, can you help me here?

zhangguanzhang:
seem kernel bug:
```
$ ip a s lo
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ee80:169:254::1/128 scope global 
       valid_lft forever preferred_lft forever
    inet6 fe81:169:254::1/128 scope link 
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
$ go run main.go fe80:169:254::1
Error listening: listen tcp6 [fe80:169:254::1]:8080: bind: invalid argument
$ go run main.go ee80:169:254::1
Listening on [ee80:169:254::1]:8080
^C
$ go run main.go fe80:169:254::1%lo
Listening on [fe80:169:254::1%lo]:8080
```

PR #420, opened 2020-11-11
check default kube-dns cm file path firstly

pacoxu:
For upgrade user, it is odd that the logs shows that /etc/kube-dns file missing.

Corefile is required, however, stubdomain is not.

So we should not set the /etc/kube-dns as the default value, we can check it and load it if it exists. 

That's why I add this pr.

### why not add it to main.go?

The kube-dns configmap mount is optional, and when we create it after the pod start, the file will appear after the config map is created.

### DIFF
--- a/cmd/node-cache/app/configmap.go
+++ b/cmd/node-cache/app/configmap.go
@@ -3,6 +3,7 @@ package app
 import (
 	"bytes"
 	"io/ioutil"
+	"os"
 	"path"
 	"strings"
 	"text/template"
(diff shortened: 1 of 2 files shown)


MrHohn (approved):
Thanks!
/lgtm
/approve

MrHohn:
cc @prameshj

cmd/node-cache/app/configmap.go
```diff
@@ -147,6 +149,12 @@ func (c *CacheApp) initDNSConfigSync() {
 	var kubeDNSChan, NodeLocalDNSChan <-chan *config.Config
 	initialKubeDNSConfig := &config.Config{}
 
+	if c.params.KubednsCMPath == "" {
+		if _, err := os.Stat(DefaultKubednsCMPath); !os.IsNotExist(err) {
```
prameshj, line 153: If kube-dns configmap is not mounted at the time that node-local-dns starts up, this will not kick off the periodic sync at all. That would be a regression in behavior.
Was the behavior verified on a kube-dns cluster?

prameshj: @pacoxu

PR #424, opened 2020-12-27
Add support for rewrite plugin in node-cache

### LINKED ISSUE #406: Support rewrite plugin for NodeLocal DNS
NodeLocal DNS does not support the CoreDNS `rewrite` plugin.  We have several customers that would like to enable NodeLocal DNS but are unable to do so because it does not support the `rewrite` plugin.

Related this, it would be ideal if there was documentation for which CoreDNS configurations are not supported for use with NodeLocal DNS.

jfeddern:
Add support for rewrite plugin in node-cache.

Fixes #406


prameshj:
travis passed - https://travis-ci.org/github/kubernetes/dns/builds/751794767

/lgtm

PR #426, opened 2020-12-28
Add coredns trace plugin to node-cache.

### LINKED ISSUE #423: Support trace plugin for NodeLocalDNS
We utilize the [trace](https://coredns.io/plugins/trace/) coredns plugin in our upstream coredns corefile as a handy way to identify when DNS requests in our kubernetes cluster are being rate limited by the native AWS VPC DNS resolvers.

I attempted to turn on the `trace` plugin in the NodeLocalDNS corefile to achieve similar visibility and was disappointed to receive the following error message:
```
[ERROR] Restart failed: no action found for directive 'trace' with server type 'dns' (missing a plugin?)
```

Is there a specific reason that some of the common CoreDNS plugins are not included in the NodeLocalDNS image?

https://github.com/kubernetes/dns/blob/master/cmd/node-cache/main.go#L19-L33

It would be nice to have access to this plugin without having to build our own NodeLocalDNS image.

iris-garcia:
Add trace coredns plugin to node-cache.

Fixes #423.

### DIFF
--- a/cmd/node-cache/main.go
+++ b/cmd/node-cache/main.go
@@ -30,6 +30,7 @@ import (
 	_ "github.com/coredns/coredns/plugin/pprof"
 	_ "github.com/coredns/coredns/plugin/reload"
 	_ "github.com/coredns/coredns/plugin/template"
+	_ "github.com/coredns/coredns/plugin/trace"
 	_ "github.com/coredns/coredns/plugin/whoami"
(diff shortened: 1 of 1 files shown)


prameshj:
Travis build passed as well - https://travis-ci.org/github/kubernetes/dns/builds/751798209

PR #427, opened 2020-12-29
Update coredns version from 1.7.0 to 1.8.0

yuzhiquan:
rt


prameshj:
Any specific feature to pick up in 1.18? I see upstream k8s is still using 1.7.
https://github.com/kubernetes/kubernetes/blob/9406e145c25e4c3041ec82529176957f164fdb08/cluster/addons/dns/coredns/coredns.yaml.in#L135

yuzhiquan:
> Any specific feature to pick up in 1.18? I see upstream k8s is still using 1.7.
> https://github.com/kubernetes/kubernetes/blob/9406e145c25e4c3041ec82529176957f164fdb08/cluster/addons/dns/coredns/coredns.yaml.in#L135

You are right,my goland ide cannot locate `_ "github.com/coredns/coredns/plugin/hosts"`,so i go get new version, may be just my ide problem.

prameshj:
> > Any specific feature to pick up in 1.18? I see upstream k8s is still using 1.7.
> > https://github.com/kubernetes/kubernetes/blob/9406e145c25e4c3041ec82529176957f164fdb08/cluster/addons/dns/coredns/coredns.yaml.in#L135
> 
> You are right,my goland ide cannot locate `_ "github.com/coredns/coredns/plugin/hosts"`,so i go get new version, may be just my ide problem.

Looks like vendor deps did not get updated, fixed it in https://github.com/kubernetes/dns/pull/430

yuzhiquan:
> > > Any specific feature to pick up in 1.18? I see upstream k8s is still using 1.7.
> > > https://github.com/kubernetes/kubernetes/blob/9406e145c25e4c3041ec82529176957f164fdb08/cluster/addons/dns/coredns/coredns.yaml.in#L135
> > 
> > 
> > You are right,my goland ide cannot locate `_ "github.com/coredns/coredns/plugin/hosts"`,so i go get new version, may be just my ide problem.
> 
> Looks like vendor deps did not get updated, fixed it in #430

Well done.
/close

PR #429, opened 2020-12-29
Trans exitchan elm type to struct

yuzhiquan:
It seems like exitchan output is useless, so struct type may be more available


prameshj:
Is there a downside to using the bool value?

yuzhiquan:
> Is there a downside to using the bool value?

image

It's likely exitChan's output data is not care, just a signal.

prameshj:
> > Is there a downside to using the bool value?
> 
> image
> 
> It's likely exitChan's output data is not care, just a signal. @prameshj

yup, understood. Either one works though.

/lgtm

PR #433, opened 2021-01-22
Increase build timeout to 1.5 hours.

prameshj:
The previous build invocation failed due to timeout - https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/dns-push-images/1352431660452286464


MrHohn (approved):
/lgtm
/approve

PR #431, opened 2021-01-20
Upgrade to dnsmasq 2.83 to pick up vulnerability fix

prameshj:
https://www.jsof-tech.com/disclosures/dnspooq

Verified that the images all come up correctly.

```
/image-checks.sh 1.16.0-18-g6f0a02c-dirty gcr.io/<repo>

1.16.0-18-g6f0a02c-dirty: Pulling from <repo>/k8s-dns-dnsmasq-nanny
Status: Downloaded newer image for gcr.io/<repo>/k8s-dns-dnsmasq-nanny:1.16.0-18-g6f0a02c-dirty
I0120 00:42:19.372813       1 main.go:78] opts: {{/usr/sbin/dnsmasq [] false} /etc/k8s/dns/dnsmasq-nanny 10000000000 127.0.0.1:10053}
I0120 00:42:19.373466       1 nanny.go:124] Starting dnsmasq []
W0120 00:42:19.390982       1 nanny.go:150] Got EOF from stdout
W0120 00:42:19.391000       1 nanny.go:150] Got EOF from stderr
F0120 00:42:19.391066       1 nanny.go:220] dnsmasq exited: <nil>


```

PR #440, opened 2021-02-18
Revert "Upgrade to dnsmasq 2.83 to pick up vulnerability fix"

prameshj:
Reverts kubernetes/dns#431

Some DNS lookup errors have been reported in http://lists.thekelleys.org.uk/pipermail/dnsmasq-discuss/2021q1/014621.html

Will pick up the latest once the fixed version is available.


MrHohn (approved):
Thanks for keeping an eye on that.
/lgtm

PR #441, opened 2021-03-04
update debian image tag 1.4.0

pacoxu:
https://github.com/kubernetes/kubernetes/pull/98526
Bump debian-base to v1.4.0 and debian-iptables to v1.5.0 

The last update is not completed. So I need to update it.


prameshj:
Can you ensure that all the images are building/running correctly? It would be great if you can attach the output of  
"./image-checks.sh <image-tag> <registry-name>"

pacoxu:
OK. I will test in my env first.

prameshj:
I see travis passed - https://travis-ci.org/github/kubernetes/dns/builds/761380512

PR #446, opened 2021-04-20
Upgrade to dnsmasq 2.85.

PurelyApplied:
Vulnerability fixed in 2.83 also introduced a DNS failure issue.
This persisted in 2.84, but should be resolved in 2.85.

See #431, #440 for context.

From [dnsmasq changelog](https://thekelleys.org.uk/dnsmasq/CHANGELOG):
> Fix problem with DNS retries in 2.83/2.84.

Pulling assignees from previous PRs.
/assign MrHohn bowei
/cc prameshj
/hold From the previous PRs, it looks like we have to give Travis some time to spin up?

### DIFF
--- a/images/dnsmasq/Makefile
+++ b/images/dnsmasq/Makefile
@@ -15,7 +15,7 @@
 VERSION ?= $(shell git describe --tags --always --dirty)
 REGISTRY ?= staging-k8s.gcr.io
 ARCH ?= amd64
-DNSMASQ_VERSION ?= dnsmasq-2.78
+DNSMASQ_VERSION ?= dnsmasq-2.85
 CONTAINER_PREFIX ?= k8s-dns
 
 ALL_ARCH := amd64 arm arm64 ppc64le s390x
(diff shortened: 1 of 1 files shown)


PurelyApplied:
I may have jumped the gun a little bit on this.  2.85 is available from https://thekelleys.org.uk/dnsmasq/, but the debian repositories listing is still 2.84-1, with 2.85-1 currently available on unstable.  

https://tracker.debian.org/pkg/dnsmasq

I'm not sure if that bothers us here, since we're pulling the tar and signature directly.

----

Also, while I'm here, [convenience link to Travis](https://travis-ci.org/github/kubernetes/dns/builds/767784567).

prameshj:
> I may have jumped the gun a little bit on this. 2.85 is available from https://thekelleys.org.uk/dnsmasq/, but the debian repositories listing is still 2.84-1, with 2.85-1 currently available on unstable.
> 
> https://tracker.debian.org/pkg/dnsmasq
> 
> I'm not sure if that bothers us here, since we're pulling the tar and signature directly.
> 
> Also, while I'm here, [convenience link to Travis](https://travis-ci.org/github/kubernetes/dns/builds/767784567).

Can we wait until 2.85 is stable? I see the current stable is 2.80.

PurelyApplied:
It looks like 2.85 has been on debian `testing` for a month now.  Did we want to wait for it to fully reach `stable`?  I noticed that the previous version we used here, 2.78, only ever reached `testing`, so I thought I'd ask.

prameshj:
Let's try this out.

/approve

PR #449, opened 2021-04-29
Avoid delaying startup by SyncInterval for initial iptables/ebtables setup

### LINKED ISSUE #448: Initial delay before iptables setup
Could you elaborate on this?

https://github.com/kubernetes/dns/blob/05a4cd0f5d3b0e7a2925ff10995eb821a683103d/cmd/node-cache/app/cache_app.go#L87-L93

Currently the default value for syncInterval is 60 seconds, which means the pod will wait for a minute before it sets up iptables for the first time. This delays the node's functional readiness, Is there a reason this may cause conflicts with other ports at that point?

I'd like the node to be ready faster, but would like syncIntervals to stay 60 seconds, should we add a flag that controls the initial delay?

eytan-avisror:
Fixes #448 

This add an initial call to setupNetworking before we start the loop in order to avoid waiting SyncInterval for initial setup.

### DIFF
--- a/cmd/node-cache/app/cache_app.go
+++ b/cmd/node-cache/app/cache_app.go
@@ -28,6 +28,7 @@ type ConfigParams struct {
 	SetupInterface       bool          // Indicates whether to setup network interface
 	InterfaceName        string        // Name of the interface to be created
 	Interval             time.Duration // specifies how often to run iptables rules check
+	InitialDelay         time.Duration // specifies the initial delay before configuring iptables
 	BaseCoreFile         string        // Path to the template config file for node-cache
 	CoreFile             string        // Path to config file used by node-cache
 	KubednsCMPath        string        // Directory where kube-dns configmap will be mounted
@@ -271,6 +272,7 @@ func (c *CacheApp) setupNetworking() {
 func (c *CacheApp) runPeriodic() {
 	c.exitChan = make(chan struct{}, 1)
 	tick := time.NewTicker(c.params.Interval * time.Second)
+	time.Sleep(c.params.InitialDelay)
 	c.setupNetworking()
 	for {
 		select {
--- a/cmd/node-cache/app/cache_app.go
+++ b/cmd/node-cache/app/cache_app.go
@@ -273,7 +273,7 @@ func (c *CacheApp) runPeriodic() {
 	// if a pidfile is defined in flags, setup iptables as soon as it's created
 	if c.params.Pidfile != "" {
 		for {
-			if isFileExist(c.params.Pidfile) {
+			if isFileExists(c.params.Pidfile) {
 				break
 			}
 			clog.Infof("waiting for coredns pidfile '%s'", c.params.Pidfile)
--- a/cmd/node-cache/app/cache_app.go
+++ b/cmd/node-cache/app/cache_app.go
@@ -84,7 +84,7 @@ func (c *CacheApp) Init() {
 	c.updateCorefile(&config.Config{})
 	// Initialize periodic sync for node-local-dns, kube-dns configmap.
 	c.initDNSConfigSync()
-	// Setup only the network interface during this init. IPTables will be setup via runPeriodic.
+	// Setup only the network interface first.
 	// This is to ensure that iptables rules don't get setup if the cache(coreDNS) is unable to startup due to config
 	// error, port conflicts or other reasons.
 	c.setupInterface()
--- a/cmd/node-cache/app/cache_app.go
+++ b/cmd/node-cache/app/cache_app.go
@@ -272,13 +272,15 @@ func (c *CacheApp) setupNetworking() {
 func (c *CacheApp) runPeriodic() {
 	// if a pidfile is defined in flags, setup iptables as soon as it's created
 	if c.params.Pidfile != "" {
-		if ok := waitForFile(c.params.Pidfile, time.Second*5, time.Second*1); !ok {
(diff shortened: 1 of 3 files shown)


cmd/node-cache/app/cache_app.go
```diff
@@ -271,6 +271,7 @@ func (c *CacheApp) setupNetworking() {
 func (c *CacheApp) runPeriodic() {
 	c.exitChan = make(chan struct{}, 1)
 	tick := time.NewTicker(c.params.Interval * time.Second)
+	c.setupNetworking()
```
prameshj, line 274: This is already called in cacheApp.Init(). Is that not sufficient?

eytan-avisror: No, because Init() skips setting up iptables and only sets up interfaces.
This means the first time we setup iptables is after the first interval of `SyncInterval`, so by default we spend 60 seconds doing nothing when we start up.

prameshj: The iptables setup was skipped in Init because we need coremain.Run() to finish before installing iptables rules to redirect traffic to node-local-dns. With the change in this PR, we could still run into that race where iptables rules got added when coreDNS was still starting up, right?

Until iptables rules are added, requests will go to the clusterDNS, so there should be no DNS downtime. How does it affect node-readiness?

eytan-avisror: I am gating my node's readiness until node-local-dns is ready to avoid workloads being scheduled during the window where the iptable rule is inserted.

If we don't want to have a race between `coremain.Run()` and setting up iptables, would it be accetable to introduce a second flag for Initial Delay which would be a more reasonable number?

My use case is that I want to keep the SyncInterval 60s but want a faster startup

eytan-avisror:
@prameshj I added this as an optional flag for initialdelay, would that be acceptable? we can change the default value to some other number if you have any preference. we would like to start up as soon as possible

cmd/node-cache/app/cache_app.go
prameshj: Instead of this, can we check for the pidfile of coreDNS to be present, to infer that coreDNS has started and then run the first instance of c.setupNetworking?

https://github.com/kubernetes/dns/blob/05a4cd0f5d3b0e7a2925ff10995eb821a683103d/vendor/github.com/caddyserver/caddy/caddy.go#L478 writes a pidfile if the "pidfile" argument is provided on the commandline.

eytan-avisror: Thanks @prameshj, this will be a more deterministic solution - I will look into it

eytan-avisror: @prameshj Done, please review.
I've done some basic functional tests and it works well, also added a unit test.
Let me know if you have any comments

cmd/node-cache/app/cache_app.go
```diff
@@ -269,6 +270,17 @@ func (c *CacheApp) setupNetworking() {
 }
 
 func (c *CacheApp) runPeriodic() {
+	// if a pidfile is defined in flags, setup iptables as soon as it's created
+	if c.params.Pidfile != "" {
+		if ok := waitForFile(c.params.Pidfile, time.Second*5, time.Second*1); !ok {
```
prameshj, line 275: If the pidfile is specified, we should block until the file is created. If it takes longer than 5s, we should still wait before starting the periodic loop, since coreDNS is still not running.

Can you also try this out on a test cluster with this image, to make sure the pidfile is indeed being created?

cmd/node-cache/app/cache_app.go
```diff
@@ -307,3 +319,29 @@ func toSvcEnv(svcName string) string {
...
+// isFileExists returns true if a file exists with the given path
+func isFileExists(path string) bool {
+	f, err := os.Stat(path)
+	if os.IsNotExist(err) {
+		return false
+	}
+	return !f.IsDir()
+}
+
+// waitForFile checks if a file exists at intervals until it file exists or a timeout is reached, returns false
```
prameshj, line 332: We might not need this function or we could just do :
```
for {
   if isFileExists(path) {
     return
   }
   time.Sleep(t)
}
```

cmd/node-cache/app/cache_app.go
eytan-avisror: I simplified the code a bit, and made the wait indefinite.
I also tested this on a cluster with and without the `pidfile` flag and it works as expected, I also exec'd into a container and verified the pidfile exist.
```
$ kubectl exec -it node-local-dns-2fhkm -- sh
# ls *.pid
coredns.pid
```

prameshj:
Travis in progress - https://travis-ci.org/github/kubernetes/dns/builds/769947702

prameshj:
oops.. PR merged before travis completed. Let us check https://k8s-testgrid.appspot.com/sig-network-dns#dns-push-images to make sure it builds correctly.

eytan-avisror:
@prameshj looks like last build went through after you increased the timeout 👍 
Thanks for your help in getting this merged - any idea when can we expect the next release that includes this?

prameshj:
1.17.4 tag has been cut. I will promote the image by end of the week.

prameshj:
> 1.17.4 tag has been cut. I will promote the image by end of the week.

Being promoted in https://github.com/kubernetes/k8s.io/pull/2047

PR #457, opened 2021-06-23
Add a new presubmit script for k/dns PRs.

prameshj:
This will be referenced by the prow job in https://github.com/prameshj/test-infra/blob/007e1d92b25e53c5474495934a732801e7a29845/config/jobs/kubernetes/sig-network/sig-network-misc.yaml#L198


MrHohn (approved):
/lgtm
/approve

PR #463, opened 2021-06-24
Add more presubmit commands to DNS Presubmit

prameshj:
This is copied over from https://github.com/kubernetes/dns/blob/master/.travis.yml, the ginkgo test is left out for now.

/assign @MrHohn


MrHohn (approved):
/lgtm
/approve

PR #464, opened 2021-06-25
Include ginkgo tests in presubmit

prameshj:
Manually verified that the ginkgo install commands work on the builder image.

/assign @MrHohn


MrHohn (approved):
/lgtm

PR #482, opened 2021-09-17
Use the right caddy dependency in node-cache 

prameshj:
coreDNS 1.7.0 uses - "github.com/caddyserver/caddy"
coreDNS 1.8.3 uses - "github.com/coredns/caddy"

When coreDNS was reverted to 1.8.3 the caddy server change was not done. This PR addresses that.

Verified that the iptables rules are cleaned up upon exit.

/assign @MrHohn


MrHohn (approved):
/lgtm

PR #483, opened 2021-09-17
Upgrade to iptables 1.6.7 to pick up vuln fixes.

prameshj:
Verified that new images start up using ./image-checks.sh

/assign @MrHohn


MrHohn (approved):
/lgtm

PR #488, opened 2021-11-05
Increment error metric for rule deletion errors as well.

prameshj:
This will help identify any teardown errors when nodelocaldns pod is exiting.

No unit test since fake iptables in https://github.com/kubernetes/kubernetes/blob/master/pkg/util/iptables/testing/fake.go does not return error in any of the calls. This needs to be enhanced to verify error scenarios.

/assign @MrHohn 
/cc @rtheis


MrHohn (approved):
/lgtm

rtheis:
Thank you.

PR #492, opened 2021-12-06
Add a new script to parse image shas for new tags.

prameshj:
This can be used to generate the diff for promoting images in https://github.com/kubernetes/k8s.io/tree/main/k8s.gcr.io/images/k8s-staging-dns

### DIFF
--- a/parse-image-sha.py
+++ b/parse-image-sha.py
@@ -1,0 +1,61 @@
+#!/usr/bin/env python
+
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


parse-image-sha.py
```diff
@@ -0,0 +1,45 @@
+# This script pulls the docker images for dns containers and outputs the sha256
```
MrHohn, line 1: Should we prepend the Apache license boilerplate here?

MrHohn (approved):
Thanks for adding this - I'm looking forward to mirror this in other repos as well :)

parse-image-sha.py
prameshj: Good idea, done.

MrHohn (approved):
/lgtm

PR #496, opened 2021-12-22
Fail presubmit if any of the build commands fail.

prameshj:
Current presubmit shows a "pass" status even if the build fails. Example - https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/dns/495/pull-kubernetes-dns-test/1473729632233590784

### DIFF
--- a/presubmits.sh
+++ b/presubmits.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+# Exit at the first failure.
+set -e
 # These are the commands run by the prow presubmit job.
 
 echo "installing ginkgo"


presubmits.sh
```diff
@@ -1,5 +1,15 @@
 #!/bin/bash
 
+function run-cmd {
```
bowei, line 3: Can we use `set -e`

https://stackoverflow.com/questions/3474526/stop-on-first-error

prameshj: Good idea, done.

PR #508, opened 2022-02-09
Use ClusterIPs field to create DNS records

dpasiukevich:
Usage of the Service.Spec.ClusterIPs field is needed for Dual-Stack clusters

### DIFF
--- a/pkg/dns/dns.go
+++ b/pkg/dns/dns.go
@@ -338,8 +338,10 @@ func (kd *KubeDNS) removeService(obj interface{}) {
 
 		// ExternalName services have no IP
 		if util.IsServiceIPSet(s) {
-			delete(kd.reverseRecordMap, s.Spec.ClusterIP)
-			delete(kd.clusterIPServiceMap, s.Spec.ClusterIP)
+			for _, ip := range util.GetClusterIPs(s) {
+				delete(kd.reverseRecordMap, ip)
+				delete(kd.clusterIPServiceMap, ip)
+			}
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


MrHohn (commented):
/lgtm

pkg/dns/util/util.go
```diff
@@ -116,3 +116,11 @@ func ValidateNameserverIpAndPort(nameServer string) (string, string, error) {
 func IsServiceIPSet(service *corev1.Service) bool {
 	return service.Spec.ClusterIP != corev1.ClusterIPNone && service.Spec.ClusterIP != ""
 }
+
+// GetClusterIPs returns IPs set for the service
+func GetClusterIPs(service *corev1.Service) []string {
+	if len(service.Spec.ClusterIPs) > 0 {
```
bowei, line 122: Can we add some defensive code and flag gate here?



if ClusterIP does not appear in ClusterIPs, let's log an error default back to using ClusterIP only.

bowei: Also -- add a test case for this..

pkg/dns/dns.go
```diff
@@ -500,13 +502,20 @@ func (kd *KubeDNS) fqdn(service *v1.Service, subpaths ...string) string {
 
 func (kd *KubeDNS) newPortalService(service *v1.Service) {
 	subCache := treecache.NewTreeCache()
-	recordValue, recordLabel := util.GetSkyMsg(service.Spec.ClusterIP, 0)
-	subCache.SetEntry(recordLabel, recordValue, kd.fqdn(service, recordLabel))
+	clusterIPs := util.GetClusterIPs(service)
+
+	for _, ip := range clusterIPs {
```
bowei, line 507: Let's factor this into its own method

bowei: inside of the for loop

dpasiukevich: To be honest, I think that the loop size and nestedness is manageable here without much need of moving the code out.
If anything I can do that in the next PR.

pkg/dns/util/util.go
dpasiukevich: > Can we add some defensive code and flag gate here?

What do you mean by defensive code here?
As for the flag gate, should we introduce command line argument to toggle the behaviour? Curious in what cases would it be needed to explicitly ignore `service.Spec.ClusterIPs` if it's present in favor of `service.Spec.ClusterIP`?

> if ClusterIP does not appear in ClusterIPs, let's log an error default back to using ClusterIP only.

If I understand correctly, logging an error here would be sort of a backwards incompatible change, e.g. if ipv4-only cluster gets this version it would just start logging errors. 
But I agree that it would probably be useful to have debug/info logs to be able quickly identify the source used to created dns entries.

> Also -- add a test case for this..

I didn't add explicit test for this func as it's invoked and the behaviour is controlled in `TestUnnamedSinglePortService` and `TestNamedSinglePortService` tests

PR #512, opened 2022-03-24
Add support for IPv6 PTR records

dpasiukevich:
Example output in dual-stack cluster.

```
root@dnsutils:/# dig +search +short my-service a my-service aaaa
10.47.32.21
2600:2d00:0:4:b1b3:daf1:6efd:bb15
root@dnsutils:/# dig +search +short -x 2600:2d00:0:4:b1b3:daf1:6efd:bb15
my-service.default.svc.cluster.local.
root@dnsutils:/# dig +search +short -x 2600:2d00::4:b1b3:daf1:6efd:bb15
my-service.default.svc.cluster.local.
root@dnsutils:/# dig +search +short -x 10.47.32.21
my-service.default.svc.cluster.local.
```


pkg/dns/util/util.go
```diff
@@ -42,14 +46,51 @@ const (
 // ExtractIP turns a standard PTR reverse record lookup name
 // into an IP address
 func ExtractIP(reverseName string) (string, bool) {
-	if !strings.HasSuffix(reverseName, ArpaSuffix) {
-		return "", false
+	if strings.HasSuffix(reverseName, ArpaSuffix) {
```
bowei, line 49: Can we change the signature to return error so we can log why something was not parsed?

pkg/dns/util/util.go
```diff
@@ -42,14 +46,51 @@ const (
...
+}
+
+// extractIPv6 turns a IPv6 PTR reverse record lookup name
+// into an IPv6 address according to RFC3596
+// b.a.9.8.7.6.5.0.4.0.0.0.3.0.0.0.2.0.0.0.1.0.0.0.0.0.0.0.1.2.3.4.ip6.arpa.
+// is reversed to 4321:0:1:2:3:4:567:89ab
+func extractIPv6(reverseName string) (string, bool) {
+	segments := ReverseArray(strings.Split(reverseName, "."))
+
+	if len(segments) != IPv6nibbleCount {
```
bowei, line 80: Does IPv6nibblecount have to be an exported const?

I would prefer const inside this func as local to the usage as possible.

pkg/dns/util/util.go
```diff
@@ -42,14 +46,51 @@ const (
...
+	if ip == nil {
+		return "", false
+	}
+	return ip.String(), true
+}
+
+// extractIPv6 turns a IPv6 PTR reverse record lookup name
+// into an IPv6 address according to RFC3596
+// b.a.9.8.7.6.5.0.4.0.0.0.3.0.0.0.2.0.0.0.1.0.0.0.0.0.0.0.1.2.3.4.ip6.arpa.
+// is reversed to 4321:0:1:2:3:4:567:89ab
```
bowei, line 76: Minor nit for accuracy: the right syntax would have the double colon (::) for the zeros that were shortened?

pkg/dns/util/util.go
```diff
@@ -119,8 +160,22 @@ func IsServiceIPSet(service *corev1.Service) bool {
 
 // GetClusterIPs returns IPs set for the service
 func GetClusterIPs(service *corev1.Service) []string {
```
bowei, line 162: I would prefer this function return actual net.IP so the normalization issue is taken care of at the closest point of use as requiring normalization here introduces a non-local dependency in the behavior of the code.

Note that technically speaking, IPv4 also has abbreviations, e.g. "1.2" is a valid IP address, but no one knows about it so it is less common.

bowei (commented):
Some comments.

pkg/dns/util/util.go
dpasiukevich: I took the ip value in func comment from RFC

https://datatracker.ietf.org/doc/html/rfc5952#section-4.2.2
```
The symbol "::" MUST NOT be used to shorten just one 16-bit 0 field.
For example, the representation 2001:db8:0:1:1:1:1:1 is correct, but
2001:db8::1:1:1:1:1 is not correct.
```

pkg/dns/util/util.go
dpasiukevich: My intention was to hide implementation details. If we would return `net.IP` that would leak to the consumers that normalization is done via `net.IP`.

If you feel there's a need to update the func to return `net.IP` I would ofc. But I'd prefer not to do that as for me the signature `func GetClusterIPs(service *corev1.Service) []string` is what we need - `GetClusterIPs` would return list of valid IPs of the service.

pkg/dns/util/util.go
dpasiukevich: done

pkg/dns/util/util.go
```diff
@@ -42,14 +46,51 @@ const (
 // ExtractIP turns a standard PTR reverse record lookup name
 // into an IP address
```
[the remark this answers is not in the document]
dpasiukevich: done

pkg/dns/util/util.go
dpasiukevich: done

pkg/dns/util/util.go
```diff
@@ -41,15 +43,64 @@ const (
...
+
+	// IPv6nibbleCount is the expected number of nibbles in IPv6 PTR record as defined in rfc3596
+	const ipv6nibbleCount = 32
+
+	if len(segments) != ipv6nibbleCount {
+		return "", fmt.Errorf("incorrect number of segments in IPv6 PTR: %v", len(segments))
+	}
+
+	var slice6 []string
+	for i := 0; i < len(segments); i += 4 {
```
kl52752, line 95: nit: you can use `ipv6nibbleCount`

pkg/dns/util/util.go
```diff
@@ -119,8 +170,22 @@ func IsServiceIPSet(service *corev1.Service) bool {
 
 // GetClusterIPs returns IPs set for the service
 func GetClusterIPs(service *corev1.Service) []string {
+	clusterIPs := []string{service.Spec.ClusterIP}
```
kl52752, line 173: can you add coment why did you added this line?

kl52752:
2 nit comments, resolve/respond to them then I can lgtm

pkg/dns/util/util.go
dpasiukevich: We iterate over segments slice so in the loop condition it's more clean to use `len(segments)` as well IMO

pkg/dns/util/util.go
dpasiukevich: I'd add the comment but I couldn't think of any useful one here.

We prefer spec.ClusterIPs over spec.ClusterIP for dual-stack support.

ClusterIPs field is a newer field and is sort of an extension for a spec.ClusterIP field (e.g. in dual-stack clusters)
And this info is already in these fields comments.

PR #529, opened 2022-07-12
Fix multi arch build and improve images/dnsmasq/validate.sh

tstapler:
The multi-arch build for the dnsmasq container was broken for ARM (and other non x86 architectures) after migrating to a distroless base in https://github.com/kubernetes/dns/commit/6c92fb0912575e1048c87b1ea51058ef940b639d. All of the dynamic dependencies included for dnsmasq were formatted for x86 machines

This PR should fix the problem by using `docker buildx` instead of `docker`.

This is the output for the bad 1.22.5 image, and the image built from my fix (It's tagged with -dirty).

New good image built locally:
```
❯ IMAGE=gcr.io/k8s-staging-dns/k8s-dns-dnsmasq-arm64:1.22.5-dirty ./validate.sh
Checking gcr.io/k8s-staging-dns/k8s-dns-dnsmasq-arm64:1.22.5-dirty 
Analyzing the binary /usr/sbin/dnsmasq from gcr.io/k8s-staging-dns/k8s-dns-dnsmasq-arm64:1.22.5-dirty
Dependencies:
libc.so.6
ld-linux-aarch64.so.1

All dependencies for usr/sbin/dnsmasq are present
WARNING: The requested image's platform (linux/arm64) does not match the detected host platform (linux/amd64) and no specific platform was requested

❯ echo $?
0
```

Old bad image:
```
❯ IMAGE=gcr.io/k8s-staging-dns/k8s-dns-dnsmasq-arm64:1.22.5 ./validate.sh
Checking gcr.io/k8s-staging-dns/k8s-dns-dnsmasq-arm64:1.22.5 
Analyzing the binary /usr/sbin/dnsmasq from gcr.io/k8s-staging-dns/k8s-dns-dnsmasq-arm64:1.22.5
Dependencies:
libc.so.6
ld-linux-x86-64.so.2

Wrong Architecture:
lib/x86_64-linux-gnu/libc-2.31.so - x86-64
lib/x86_64-linux-gnu/ld-2.31.so - x86-64

Missing Dependencies:
ld-linux-aarch64.so.1

❯ echo $?
1
```

### DIFF
--- a/images/dnsmasq/validate.sh
+++ b/images/dnsmasq/validate.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -e
 
 if [ -z "$IMAGE" ]; then
   echo "IMAGE needs to be set to the dnsmasq image"
@@ -6,5 +7,10 @@ fi
 
 echo "Checking ${IMAGE} ${ARCH}"
 
+# Check that dnsmasq is correctly compiled and has the right
+# dynamic libraries to run.
+python3 validate_dynamic_deps.py --image ${IMAGE} --target-bin /usr/sbin/dnsmasq
+
 # Check that dnsmasq is able to start.
 exec docker run --rm -- "${IMAGE}" --version >/dev/null
+
--- a/images/dnsmasq/validate_dynamic_deps.py
+++ b/images/dnsmasq/validate_dynamic_deps.py
@@ -1,0 +1,189 @@
+#/usr/bin/env python3
+
+# Copyright 2022 The Kubernetes Authors.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+# http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# This script requires python 3.8+
+import os
+import shutil
+import subprocess
+import pathlib
+import sys
+from typing import List
+
+IMAGE_DIR = '.test-images'
+
+
+def replace_img_special_chars(image_ref: str) -> str:
+    """Replace container image name characters that cannot be a part of container names or file names."""
+    return image_ref.lower().replace('/', '_').replace(':', '_')
+
+
+def relative_to_root(path: pathlib.Path) -> pathlib.Path:
+    return path.relative_to(path.anchor)
+
+
+def detect_architecture(cwd: pathlib.Path, target_binary: pathlib.Path) -> str:
+    """Read the architecture of the ELF file target_binary using the file command"""
+    result = subprocess.run(['file', target_binary],
+                            capture_output=True,
+                            text=True,
+                            cwd=cwd)
+    return result.stdout.split(",")[1].strip()
+
+
+def untar_container_image(image_ref: str, output_dir: pathlib.Path):
+    """Create a docker container using image_ref use tar to extract the filesystem to output_dir."""
+    container_name = replace_img_special_chars(image_ref)
+    subprocess.run(['docker', 'create', '--name', container_name, image_ref],
+                   capture_output=True)
+    subprocess.run(f'docker export {container_name} | tar x',
+                   shell=True,
+                   cwd=output_dir,
+                   capture_output=True)
+    subprocess.run(['docker', 'rm', container_name], capture_output=True)
+
+
+def detect_dependencies(cwd: pathlib.Path,
+                        target_binary: pathlib.Path) -> [str]:
+    """Use objdump to read the ELF headers for target_binary and return a
+    list of .so files.
+
+    objdump is used instead of ldd because it reads the binary's contents
+    instead of trying to run the binary. This means that it can be used on
+    binaries with different architectures than the host machine.
+
+    Using ldd on an x86 machine against x86 binary:
+    ❯ ldd arm-container-test/k8s-dns-dnsmasq-amd64-1.22.3/usr/sbin/dnsmasq
+        linux-vdso.so.1 (0x00007ffc3b9f5000)
+        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fb2f44f8000)
+        /lib64/ld-linux-x86-64.so.2 (0x00007fb2f4748000)
+
+    Using ldd x86 machine against an ARM binary:
+    ❯ ldd arm-container-test/k8s-dns-dnsmasq-arm64-1.22.3/usr/sbin/dnsmasq
+        not a dynamic executable
+    """
+    result = subprocess.run(['objdump', '--private-headers', target_binary],
+                            capture_output=True,
+                            text=True,
+                            cwd=cwd)
+    deps = []
+    for line in result.stdout.splitlines():
+        if "NEEDED" in line:
+            deps.append(line.split()[1])
+    return deps
+
+
+def find_dependencies_with_name(root_dir: pathlib.Path, dependency_name: str) -> \
+    List[pathlib.Path]:
+    """Search for the first file called dependency_name within root_dir. """
+    return list(root_dir.glob(f'**/{dependency_name}'))
+
+
+def resolve_container_link(root_dir: pathlib.Path,
+                           link: pathlib.Path) -> pathlib.Path:
+    """Resolve the provided link, make absolute paths relative to root_dir."""
+    resolved = os.readlink(link)
+
+    if str(resolved).startswith('/'):
+        return root_dir.joinpath(relative_to_root(resolved))
+    return link.resolve().relative_to(root_dir.resolve())
+
+
+def main(image_ref: str, target_binary: str, clean=True):
+    print(f"Analyzing the binary {target_binary} from {image_ref}")
+    container_name = replace_img_special_chars(image_ref)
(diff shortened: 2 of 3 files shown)


tstapler:
FYI @dpasiukevich

images/dnsmasq/validate_dynamic_deps.sh
```diff
@@ -0,0 +1,113 @@
...
+if [ ${#missing[@]} -gt 0 ]; then
+echo Missing Dependencies:
+for miss in "${missing[@]}"; do
+  echo "$miss"
+done
+exit_code=1
+else
+  echo "All dependencies for $BIN_PATH are present in $IMAGE_CONTENT_DIR"
+fi
+
```
dpasiukevich, line 112: we should clean up $IMAGE_CONTENT_DIR after the run

images/dnsmasq/validate_dynamic_deps.sh
```diff
@@ -0,0 +1,113 @@
...
+  if [[ "$current_location" == "" ]]; then
+    seen[$current]=true
+    missing+=( "$current" )
+    # Remove first element
+    dependencies=( "${dependencies[@]:1}" )
+    continue
+  fi
+
+  # Expand symlinks because objdump doesn't follow them
+  while [ -L "$current_location" ]; do
```
dpasiukevich, line 62: TBH Would it be a good idea to have a python script instead of bash? It a little effort but that would be a major readability/supportability boost.

E.g. chain expand symlinks in python would be just: `os.path.realpath(current_location)`

tstapler: Yep, I'll rewrite this in python real quick.

images/dnsmasq/validate.sh
```diff
@@ -1,10 +1,22 @@
...
 fi
 
 echo "Checking ${IMAGE} ${ARCH}"
 
+# Check that dnsmasq is correctly compiled and has the right
+# dynamic libraries to run.
+"$SCRIPT_DIR/validate_dynamic_deps.sh" ${IMAGE} /usr/sbin/dnsmasq
+exit_code=$?
+
+if [ $exit_code -ne 0 ]; then
```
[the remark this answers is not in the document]
tstapler: Sounds good

images/dnsmasq/validate_dynamic_deps.sh
```diff
@@ -0,0 +1,113 @@
...
+# are included in the image.
+#
+# Example:
+# ./verify_dynamic_deps.sh gcr.io/k8s-staging-dns/k8s-dns-dnsmasq-arm64:1.22.5 /usr/sbin/dnsmasq
+
+TARGET_IMAGE=$1
+CONTAINER_NAME=$(sed -r --expression "s%[^/]+/([^/]+)/([^:]+):(.*)%\2%" <<< "$TARGET_IMAGE")
+TAG=$(sed -r --expression "s%[^\/]+/([^\/]+)/([^:]+):(.*)%\3%" <<< "$TARGET_IMAGE")
+
+IMAGE_CONTENT_DIR="test-images/$CONTAINER_NAME-$TAG"
```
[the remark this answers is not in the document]
tstapler: Avoided using this gnarly regex in the python version.

images/dnsmasq/validate_dynamic_deps.sh
```diff
@@ -0,0 +1,113 @@
...
+mkdir -p "$IMAGE_CONTENT_DIR"
+(cd "$IMAGE_CONTENT_DIR" &&
+docker create --name "$CONTAINER_NAME" $TARGET_IMAGE > /dev/null && \
+(docker export $CONTAINER_NAME | tar x) && docker rm $CONTAINER_NAME >/dev/null)
+
+find_arch() {
+  echo "$(file "$1" | awk -F',' '{print $2}')"
```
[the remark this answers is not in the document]
tstapler: awk is no longer required in the python version.

images/dnsmasq/validate.sh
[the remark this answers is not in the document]
tstapler: Sounds good, made this change

images/dnsmasq/validate.sh
```diff
@@ -1,10 +1,22 @@
...
 
 if [ -z "$IMAGE" ]; then
   echo "IMAGE needs to be set to the dnsmasq image"
 fi
 
 echo "Checking ${IMAGE} ${ARCH}"
 
+# Check that dnsmasq is correctly compiled and has the right
+# dynamic libraries to run.
+"$SCRIPT_DIR/validate_dynamic_deps.sh" ${IMAGE} /usr/sbin/dnsmasq
```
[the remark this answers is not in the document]
tstapler: Sure, made this change.

images/dnsmasq/validate_dynamic_deps.py
```diff
@@ -0,0 +1,158 @@
+#/usr/bin/env python3
```
bowei, line 1: copyright

is this code copied from somewhere else?

tstapler: Just to confirm, you're saying to add the Kubernetes copyright?

This is original code.

bowei: Yes, you need the copyright header.

tstapler: Added the copyright header

images/dnsmasq/validate_dynamic_deps.py
```diff
@@ -0,0 +1,170 @@
...
+                   shell=True,
+                   cwd=output_dir,
+                   capture_output=True)
+    subprocess.run(['docker', 'rm', container_name], capture_output=True)
+
+
+def detect_dependencies(cwd: pathlib.Path,
+                        target_binary: pathlib.Path) -> [str]:
+    """Use objdump to read the ELF headers for target_binary and return a list of .so files."""
+    result = subprocess.run(['objdump', '--private-headers', target_binary],
```
dpasiukevich, line 58: Why `objdump` is used?
IIUC `ldd target_binary` would give you a full dependency graph with filepath of those dependencies

tstapler: `objdump` is used because it reads the binary's. `ldd` will actually try to run the binary using a special `LD_TRACE_LOADED_OBJECTS` flag which gets the dependencies.

Running ldd on my x86 machine against the x86 container
```shell
❯ ldd arm-container-test/k8s-dns-dnsmasq-amd64-1.22.3/usr/sbin/dnsmasq
        linux-vdso.so.1 (0x00007ffc3b9f5000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fb2f44f8000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fb2f4748000)
```

Running ldd on my x86 machine against the ARM container

```shell
❯ ldd arm-container-test/k8s-dns-dnsmasq-amd64-1.22.3/usr/sbin/dnsmasq
        linux-vdso.so.1 (0x00007ffc3b9f5000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fb2f44f8000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fb2f4748000)
```

images/dnsmasq/validate_dynamic_deps.py
```diff
@@ -0,0 +1,170 @@
...
+    deps = []
+    for line in result.stdout.splitlines():
+        if "NEEDED" in line:
+            deps.append(line.split()[1])
+    return deps
+
+
+def find_dependency_path(root_dir: pathlib.Path, dependency_name: str)-> Optional[pathlib.Path]:
+    """Search for the first file called dependency_name within root_dir."""
+    paths = sorted(root_dir.glob(f'**/{dependency_name}'))
```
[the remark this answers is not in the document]
tstapler: Shared object files are resolved by name, so we should only need the first file out of the list. This will not work if the container had multiple copies of the required `.so` files.

While I was looking into less naive ways of resolving these dependencies ,I noticed that the ldconfig tool has `-r` option to use the container's fs as a root. We can use it to load all of the .so files in the image. I'll update the script to use that instead. it should be more reliable.

images/dnsmasq/validate_dynamic_deps.py
tstapler: I added a comment to the function explaining why it uses objdump

images/dnsmasq/validate_dynamic_deps.py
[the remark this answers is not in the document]
tstapler: `ldconfig` worked on multiple architectures, so I used it to replace this function. Let me know what you think of the new dependency resolution strategy.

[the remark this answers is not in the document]
tstapler: Nevermind, ldconfig can read cache files generated on other architectures, but it doesn't when generating new cache files.

I'll revert back to the globbing which I was doing before and make it handle multiple `.so` files with the same name.
```
❯ ldconfig -r .test-images/gcr.io_k8s-staging-dns_k8s-dns-dnsmasq-arm64_1.22.5-dirty -v   
ldconfig: Can't stat /lib/x86_64-linux-gnu: No such file or directory
ldconfig: Can't stat /usr/lib/x86_64-linux-gnu: No such file or directory
/lib: (from <builtin>:0)
ldconfig: /lib/ld-linux-aarch64.so.1 is for unknown machine 183.

/usr/lib: (from <builtin>:0)
```

tstapler:
Latest changes to use `ldconfg` based off of feedback https://github.com/kubernetes/dns/pull/529#discussion_r921356066 don't work in all cases. 

Reverting back to globbing

tstapler:
Moved the script to using globbing again. Updated it to handle duplicate dependencies .

PR #532, opened 2022-07-18
Update cloudbuild.yaml to run docker buildx

dpasiukevich:
In README https://github.com/kubernetes/test-infra/blob/b74506333229b4cecfddccded988e80f5742e4b3/config/jobs/image-pushing/README.md

it's said to set HOME var for the docker buildx build

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -9,6 +9,7 @@ steps:
     - PROW_GIT_TAG=$_GIT_TAG
     - PULL_BASE_REF=$_PULL_BASE_REF
     - VERSION=$_PULL_BASE_REF
+    - HOME=/root  # for docker buildx
     args:
     - -c
     - |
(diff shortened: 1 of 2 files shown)


cloudbuild.yaml
```diff
@@ -9,6 +9,7 @@ steps:
     - PROW_GIT_TAG=$_GIT_TAG
     - PULL_BASE_REF=$_PULL_BASE_REF
     - VERSION=$_PULL_BASE_REF
+    - HOME=/root  # for docker buildx
```
prameshj, line 12: Looks like this is needed only if we use docker buildx. Is that the case?

dpasiukevich: `buildx build` was used for dnsmasq image. And I expected that changing homedir should work for both cases (using buildx and build)

But I've updated `rules.mk` to use buildx as well for consistency.

Issue #539, opened 2022-08-23
`node-local-dns` requires extra validations for `kube-dns` configmap


tangenti:
`node-local-dns` loads the values of `stubDomains` and `upstreamNameservers` from `kube-dns` configmap and compiles the config into a Corefile. However, a valid `kube-dns` config may be compiled into an invalid Corefile.

For example, the following config would be considered as valid `kube-dns` config.
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: kube-dns
  namespace: kube-system
data:
  stubDomains: |
        {"example.com" : ["cluster.local"]}
  upstreamNameservers: |
        ["8.8.4.4"]
```

The name servers specified in the `stubDomains` fields will then be used as name servers for the plugin `forward` in the compiled Corefile. However, `forward` does not support FQDNs for the name servers, and it crashes when it sees such a configuration, and the container falls into a crash loop. Therefore we will need extra validations to make sure the values are valid for the `forward` plugin.

chrisohaver:
> following config would be considered as valid kube-dns config.

How does kube-dns behave with such a config?  Does it try to configure dnsmasq with a domain as an upstream ip?  Is that allowed in dnsmasq?

tangenti:
> > following config would be considered as valid kube-dns config.
> 
> How does kube-dns behave with such a config? Does it try to configure dnsmasq with a domain as an upstream ip? Is that allowed in dnsmasq?

Hi Chris, I'm not an expert of `kube-dns`. Checking the source code https://github.com/tangenti/dns/blob/bfa8ca85b4bb1ffcadba651500b504818f04d3ce/pkg/dnsmasq/nanny.go#L83, it seems the name servers will be resolved to an IP address first.

PR #540, opened 2022-08-23
Add extra config validation in node-local-cache config sync

tangenti:
The validation is ported from the CoreDNS plugin forward: https://github.com/coredns/coredns/blob/fc7f3835d506960364b21cf62116a70eb8f4caf2/plugin/forward/setup.go#L97

TESTED= make bulid; make containers; make test (with the broken validate_dynamic_deps.py skipped)
Fix: #539

### DIFF
--- a/cmd/node-cache/app/cache_app_test.go
+++ b/cmd/node-cache/app/cache_app_test.go
@@ -246,7 +246,7 @@ func TestUpdateIPv6CoreFile(t *testing.T) {
 	}
 	customConfig := &config.Config{StubDomains: map[string][]string{
 		"acme.local":   {"2001:db8:1:1:1::1"},
-		"google.local": {"8.8.8.8"},
+		"google.local": {"2001:4860:4860::8888"},
 		"widget.local": {"[2001:db8:2:2:2::2]:10053", "2001:db8:3:3:3::3"},
 	},
 		UpstreamNameservers: []string{"[2001:db8:2:2:2::2]:10053", "2001:db8:3:3:3::3"},
--- a/cmd/node-cache/app/cache_app_test.go
+++ b/cmd/node-cache/app/cache_app_test.go
@@ -170,7 +170,7 @@ func TestUpdateCoreFile(t *testing.T) {
 	}
 	customConfig := &config.Config{StubDomains: map[string][]string{
 		"acme.local":   {"1.1.1.1"},
-		"google.local": {"google-public-dns-a.google.com"},
+		"google.local": {"8.8.8.8"},
 		"widget.local": {"2.2.2.2:10053", "3.3.3.3"},
 	},
 		UpstreamNameservers: []string{"2.2.2.2:10053", "3.3.3.3"},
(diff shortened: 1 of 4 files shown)


tangenti:
CC @dpasiukevich

pkg/dns/config/config.go
```diff
@@ -130,3 +132,39 @@ func (config *Config) validateUpstreamNameserver() error {
...
+		return nil
+	}
+	hosts, err := parse.HostPortOrFile(nameservers...)
+	if err != nil {
+		return err
+	}
+	for _, host := range hosts {
+		trans, _ := parse.Transport(host)
+		switch trans {
+		case "dns", "tls":
```
dpasiukevich, line 164: As the coredns version used is not the latest AFAIK https://github.com/kubernetes/dns/blob/1fb1456f536245a6fd09622b84e6a7e918287f00/go.mod#L7

Could you verify that "dns"/"tls" protocols are supported in that older coredns version. So that we can allow it here in out validation?

dpasiukevich (commented):
The changes are ok with a bit of minor suggestions. Thanks for the PR!

pkg/dns/config/config.go
tangenti: Yes, it's also valid in the older version: https://github.com/coredns/coredns/blob/v1.7.0/plugin/forward/setup.go#L101

cmd/node-cache/app/cache_app_test.go
```diff
@@ -246,7 +246,7 @@ func TestUpdateIPv6CoreFile(t *testing.T) {
 	}
 	customConfig := &config.Config{StubDomains: map[string][]string{
 		"acme.local":   {"2001:db8:1:1:1::1"},
-		"google.local": {"google-public-dns-a.google.com"},
+		"google.local": {"8.8.8.8"},
```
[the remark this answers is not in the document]
tangenti: Done.

pkg/dns/config/config.go
```diff
@@ -21,6 +21,8 @@ import (
 	"net"
 	"strconv"
 
+	"github.com/coredns/coredns/plugin/pkg/parse"
+
```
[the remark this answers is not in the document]
tangenti: Done.

pkg/dns/config/config.go
```diff
@@ -130,3 +132,39 @@ func (config *Config) validateUpstreamNameserver() error {
 	}
 	return nil
 }
+
+// ValidateNodeLocalCacheConfig returns nil if the config can be compiled
+// to a valid Corefile.
+func (config *Config) ValidateNodeLocalCacheConfig() error {
+	for _, stubDomains := range config.StubDomains {
+		if err := validateForwardProxy(stubDomains...); err != nil {
+			return err
```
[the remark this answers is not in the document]
tangenti: Done.

pkg/dns/config/config.go
```diff
@@ -130,3 +131,39 @@ func (config *Config) validateUpstreamNameserver() error {
...
+// ValidateNodeLocalCacheConfig returns nil if the config can be compiled
+// to a valid Corefile.
+func (config *Config) ValidateNodeLocalCacheConfig() error {
+	for domain, nameservers := range config.StubDomains {
+		if err := validateForwardProxy(nameservers...); err != nil {
+			return fmt.Errorf("invalid nameservers %s for the stub domain %s: %v", nameservers, domain, err)
+		}
+	}
+	if err := validateForwardProxy(config.UpstreamNameservers...); err != nil {
+		return err
```
dpasiukevich, line 144: could you also add context to the error here similar to stub-domains? thanks!

tangenti: Oops. Done!

PR #551, opened 2022-11-09
Imported and upgraded skydns dependency (plus upgraded coredns to the latest 1.10.0)

dpasiukevich:
With coredns 1.10.0 I don't observe regression for the request misses.

image

First graph is total requests
Second graph shows volume of the requests that are forwarded to the upstream.

There's 6000/300 ratio of all requests to nodelocaldns and forwarded requests. Original issue https://github.com/kubernetes/dns/issues/476 with the regression shows 5k/1.2k


kl52752 (approved):

SuperQ:
Nice, thanks for getting CoreDNS updated.

Issue #564, opened 2023-01-05
k8s-dns-node-cache:1.22.16 image missing


johngmyers:
Noticed the GitHub release announcement of 1.22.16. Tried to upgrade. The image is not in the expected place:

```
% crane manifest registry.k8s.io/dns/k8s-dns-node-cache:1.22.16
Error: fetching manifest registry.k8s.io/dns/k8s-dns-node-cache:1.22.16: GET https://us-west2-docker.pkg.dev/v2/k8s-artifacts-prod/images/dns/k8s-dns-node-cache/manifests/1.22.16: MANIFEST_UNKNOWN: Failed to fetch "1.22.16"
```

/kind bug

dpasiukevich:
The build https://testgrid.k8s.io/sig-k8s-infra-gcb#dns-push-images

* passed on main branch merge - https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/dns-push-images/1610587357218607104
* timed out on tag release - https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/dns-push-images/1610620874375827456

Both run on the same commit https://github.com/kubernetes/dns/commit/13d798c231c881510a0eb88b2a917a40f996a550

@BenTheElder, could you increase the timeout value in the config to avoid similar flakes in the future? And rerun the build (as I don't have permissions for that)

BenTheElder:
The timeout is configured in the sources / config files. I did trigger a rerun.


https://github.com/kubernetes/dns/blob/13d798c231c881510a0eb88b2a917a40f996a550/cloudbuild.yaml#L4 is well below the 2h default timeout we have set in the prowjob, you can set the timeout higher in both the cloudbuild and prow, but right now it's hitting the cloudbuild timeout.

dpasiukevich:
Thanks, @BenTheElder!

> is well below the 2h default timeout we have set in the prowjob
1. Just to clarify, 7200sec is 2 hours. So if I increase the timeout in the cloudbuild.yaml to be >7200sec, won't it hit 2h default timeout of the prowjob?
2. Any maybe you know, if CI job builds the 1.22.16 tag, which cloudbuild.yaml file will be applied: from 1.22.16 tag or from the main branch?

BenTheElder:
1. Seems I dropped a zero :+), It will, you'll need to increase both. `timeout` field on the prowjob, with a go time.Duration string. (the prow docs are in flux at the moment moving to a new site ...)
[...]

PR #565, opened 2023-01-08
Increase build timeout

dpasiukevich:
The build for all images for all supported arch takes ~1h50min and it may not fit the current 2h sometimes.  https://github.com/kubernetes/dns/issues/564#issuecomment-1372030269

One of the options is to investigate if the build parallelisation would decrease the build time, but that's not a priority and may be would be done later.


kl52752:
\lgtm
\approve

PR #571, opened 2023-02-03
Update DNS specification with RFC1035

aojea:
Maybe is redundant, but for compatibility reasons we don't want Kubernetes DNS be different than the other DNS implementations, so all the same rules and limitations apply.

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -8,6 +8,8 @@ other protocols and mechanisms, DNS is very commonly used and is a
 highly recommended add-on. The actual DNS service itself need not
 be provided by the default Kube-DNS implementation. This document
 is intended to provide a baseline for commonality between implementations.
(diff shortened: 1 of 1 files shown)


chrisohaver:
I think this may be too broad in this context (K8s Service Discovery).

Are there any DNS standard behaviors in particular that have been problematic or confusing in existing implementations (e.g. CoreDNS and Kube-DNS)?

thockin:
I don't understand what this means, in practice.  That they must be at least 1035?  What about subsequent specs?  Do we have any evidence that this is a problem for any implementation?

aojea:
nevermind, this came up with a discussion about the max number of DNS records , and some argumente was  "project X supports Y records" and Y was out of the spec.

But it turns out that the data  was wrong, Y was not the right number ... and I forgot to close this

/close

PR #577, opened 2023-02-16
Update dependencies

wewark:
Update dependencies:
- go version 1.18 -> 1.20
- golang.org/x/net 0.1.0 -> 0.7.0
- k8s.io/kubernetes 1.24.7 -> 1.24.10

Remove deprecated `-i` flag as it was causing weird behavior resulting in test error. As per golang [release notes](https://go.dev/doc/go1.16#go-command):
> The -i flag accepted by go build, go install, and go test is now deprecated. The -i flag instructs the go command to install packages imported by packages named on the command line. Since the build cache was introduced in Go 1.10, the -i flag no longer has a significant effect on build times, and it causes errors when the install directory is not writable.

Fix fmt for some files that was apparently introduced by new fmt rules in go1.20.


wewark:
PR description updated.

PR #552, opened 2022-11-09
Upgraded to the latest golang crypto package for vuln fixes

dpasiukevich:
Fixes https://github.com/kubernetes/dns/issues/522

PR #599, opened 2023-09-12
upgrate go to 1.20.7 and remove hardcoded crypto version

kl52752:
fixes for Vulnerabilities: CVE-2023-29409, CVE-2023-39533, CVE-2023-29406

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -115,12 +115,9 @@ require (
 
 replace (
 	// pinned latest version for vulnerability fixes
-	// this one is used by coredns
+    // this one is used by coredns
 	// if coredns starts using >= v0.14.0 this pinned version can be removed
 	github.com/apache/thrift => github.com/apache/thrift v0.14.0
-
-	// pinned latest version for vulnerability fixes, upgrade if there are newer versions
-	golang.org/x/crypto => golang.org/x/crypto v0.1.0
 
 	k8s.io/api => k8s.io/api v0.24.7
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -115,13 +115,10 @@ require (
 
 replace (
 	// pinned latest version for vulnerability fixes
-	// this one is used by coredns
+    // this one is used by coredns
 	// if coredns starts using >= v0.14.0 this pinned version can be removed
 	github.com/apache/thrift => github.com/apache/thrift v0.14.0
 
-	// pinned latest version for vulnerability fixes, upgrade if there are newer versions
-	golang.org/x/crypto => golang.org/x/crypto v0.1.0
```
dpasiukevich, line 123: This is a good cleanup!

Just to ensure -- did you verify:
1. This builds correctly.
2. `golang.org/x/crypto` is not used at all or the version >= v0.1.0 is used? Not to regress the https://github.com/kubernetes/dns/pull/552

kl52752: 1. yes
2. From what I understand if makefile has 1.20.7 version hardcoded then this version of crypto library will be used. DIMS?

dpasiukevich: My expectation, although the golang version that used to build the binary is updated, the packages version will be used from go.sum.
So I'd expect the binary would not contain `golang.org/x/crypto` at all.

If that's fine, could you run `go version -m`? We'd ensure that the dependencies are correct and merge the PR.

kl52752: verified, there is no dependency on crypto library

PR #602, opened 2023-10-19
Configure Dependabot for automatic dependency updates

zhuxiaow0:
Enable Dependabot for Dockerfile and go.mod updates.
This will update images in Dockerfile `FROM` statements, and go modules.

The purpose is to drastically reduce number of vulnerabilities that affect this repo, 
by keeping dependencies up-to-date. 

This PR is created by a script. Please let me know if we should modify any values.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,18 @@
+version: 2
+updates:
+- package-ecosystem: docker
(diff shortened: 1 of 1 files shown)


MrHohn (approved):
/lgtm
/approve

cc @jingyuanliang @kl52752

Issue #605, opened 2023-10-24
pull-kubernetes-dns-test broken at head


jingyuanliang:
Found in: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/dns/603/pull-kubernetes-dns-test/1715355423860592640

Confirmed failure without code change at all: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/dns/604/pull-kubernetes-dns-test/1716792265403273216 (1716792265403273216	#604 (eaaed71) by jingyuanliang 08:22:35 26m57s	FAILURE)

Manipulated test code: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/dns/604/pull-kubernetes-dns-test/1716834170107334656 (1716834170107334656	#604 (68b664b) by jingyuanliang 11:09:05 26m59s	SUCCESS)

And extracted the failure:

```
2023/10/24 15:35:43 [1mLOG: [0mdocker | Flag --api-servers has been deprecated, Use --kubeconfig instead. Will be removed in a future version.
2023/10/24 15:35:43 [1mLOG: [0mdocker | Flag --config has been deprecated, Use --pod-manifest-path instead. Will be removed in a future version.
2023/10/24 15:35:43 [1mLOG: [0mdocker | I1024 15:35:22.119199  199402 feature_gate.go:181] feature gates: map[]
2023/10/24 15:35:43 [1mLOG: [0mdocker | W1024 15:35:22.119532  199402 server.go:605] Could not load kubeconfig file /var/lib/kubelet/kubeconfig: stat /var/lib/kubelet/kubeconfig: no such file or directory. Using default client config instead.
2023/10/24 15:35:43 [1mLOG: [0mdocker | I1024 15:35:22.665666  199402 server.go:536] Using self-signed cert (/var/run/kubernetes/kubelet.crt, /var/run/kubernetes/kubelet.key)
2023/10/24 15:35:43 [1mLOG: [0mdocker | I1024 15:35:22.665796  199402 server.go:124] Running kubelet in containerized mode (experimental)
2023/10/24 15:35:43 [1mLOG: [0mdocker | I1024 15:35:22.665986  199402 docker.go:356] Connecting to docker on unix:///var/run/docker.sock
2023/10/24 15:35:43 [1mLOG: [0mdocker | I1024 15:35:22.666057  199402 docker.go:376] Start docker client with request timeout=2m0s
[...]

PR #606, opened 2023-10-24
Don't run kubelet in the E2E test mock cluster

jingyuanliang:
The existing E2E tests don't need kubelet (proven as this commit still
passing presubmit), while the execution of kubelet started to cause
mysterious failures recently (#605). Let's exclude kubelet in the mock
cluster for now. Not completely removing the relevant code in case some
future tests need them, then #605 can be investigated at that time.


bowei:
Can we have more details on the commit message on why this change is being done?

jingyuanliang:
/hold
Just pushed to run the test. Will amend for commit messages etc.

PR #610, opened 2023-11-03
Use tag gcr.io/distroless/base-debian11:latest for dnsmasq base image

jingyuanliang:
To alleviate the pain of continuous bumping.

### DIFF
--- a/images/dnsmasq/Makefile
+++ b/images/dnsmasq/Makefile
@@ -28,8 +28,7 @@ OUTPUT_DIR := _output/$(ARCH)
 export DOCKER_CLI_EXPERIMENTAL=enabled
 
 # Multiarch image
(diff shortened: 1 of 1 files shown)


images/dnsmasq/Makefile
```diff
@@ -29,7 +29,7 @@ export DOCKER_CLI_EXPERIMENTAL=enabled
 
 # Multiarch image
 # Uploaded: Jun 2, 2023, 12:57:53 PM
```
DamianSawicki, line 31: I believe this line should be removed if we are to switch from a specific sha to `latest`.

jingyuanliang: done

bowei:
Please rebase?

PR #612, opened 2023-11-06
Bump golang to 1.21.3-bookworm to fix vulnerabilities

DamianSawicki:
/assign @kl52752

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ REGISTRY ?= gcr.io/k8s-staging-dns
 # Default architecture to build for.
 ARCH ?= amd64
 # Image to use for building.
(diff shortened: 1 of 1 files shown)


jingyuanliang:
Not necessary to do this. Just rebuild.

DamianSawicki:
Ok, it is indeed not necessary to bump because the current tag `golang:1.21-bookworm` points to the same as the proposed tag `golang:1.21.3-bookworm` (see [here](https://github.com/docker-library/official-images/blob/1725ca745abc0f6483a2f007bde4cfd5014cc67d/library/golang#L8)), so if we build now there'll be no difference, and in the future the current tag will keep pointing to the latest patch version.

PR #623, opened 2024-02-14
Upgrade to dnsmasq v2.90

DamianSawicki:
The new dnsmasq version is officially released. Among other things, it 
 * enables increasing the number of concurrent TCP connections, previously hardcoded as 20,
 * fixes the bug causing memory leaks for records with TTL=0.


aojea:
do we want to use a larger number of concurrent tcp connections by default?

bowei:
Can we do a small experiment to understand if we can raise it further (TCP)

DamianSawicki:
> do we want to use a larger number of concurrent tcp connections by default?

My plan is to upgrade the image first and only later add the flag.

PR #644, opened 2024-10-06
Send SIGUSR1 to dnsmasq periodically

DamianSawicki:
This adds a feature flag `--logInterval` for periodic triggering dnsmasq to log its statistics by sending `SIGUSR1` to it.

The new motivation for adding this feature comes from dnsmasq changes in `v2.90`: the new [flag](https://thekelleys.org.uk/gitweb/?p=dnsmasq.git;a=commit;h=416390f9962e455769aa8ab6df0e105cae07ae55) `--max-tcp-connections`and the related [statistics](https://thekelleys.org.uk/gitweb/?p=dnsmasq.git;a=commit;h=69877f565a969260cf2761b2fd512d6c28dbf2f0) of TCP connection utilisation added to the log output triggered by `SIGUSR1`.

The change was benchmarked (see the commit description) and no negative impact was observed.

### DIFF
--- a/cmd/dnsmasq-nanny/main.go
+++ b/cmd/dnsmasq-nanny/main.go
@@ -37,6 +37,7 @@ var (
 		RunNannyOpts: dnsmasq.RunNannyOpts{
 			DnsmasqExec:     "/usr/sbin/dnsmasq",
 			RestartOnChange: false,
+			LogInterval:     time.Duration(0),
 		},
 		configDir:     "/etc/k8s/dns/dnsmasq-nanny",
 		syncInterval:  10 * time.Second,
@@ -69,6 +70,9 @@ Any arguments given after "--" will be passed directly to dnsmasq itself.
 		"interval to check for configuration updates")
 	flag.StringVar(&opts.kubednsServer, "kubednsServer", opts.kubednsServer,
 		"local kubedns instance address for non-IP name resolution")
+	flag.DurationVar(&opts.LogInterval, "logInterval",
+		opts.LogInterval,
+		"interval to send SIGUSR1 to dnsmasq which triggers statistics logging (if zero, SIGUSR1 is not sent)")
 	klog.InitFlags(nil)
 	flag.Parse()
 }
--- a/pkg/dnsmasq/nanny.go
+++ b/pkg/dnsmasq/nanny.go
@@ -176,7 +176,6 @@ func (n *Nanny) Kill() error {
 	}
 
 	if err := n.cmd.Process.Kill(); err != nil {
-		klog.Errorf("Error killing dnsmasq: %v", err)
 		return err
 	}
 
@@ -222,10 +221,14 @@ func RunNanny(sync config.Sync, opts RunNannyOpts, kubednsServer string) {
 		case currentConfig = <-configChan:
 			if opts.RestartOnChange {
 				klog.V(0).Infof("Restarting dnsmasq with new configuration")
-				nanny.Kill()
+				if err := nanny.Kill(); err != nil {
+					klog.Errorf("Error killing dnsmasq: %v", err)
+				}
 				nanny = &Nanny{Exec: opts.DnsmasqExec}
 				nanny.Configure(opts.DnsmasqArgs, currentConfig, kubednsServer)
-				nanny.Start()
+				if err := nanny.Start(); err != nil {
+					klog.Errorf("Could not start dnsmasq with new configuration: %v", err)
+				}
 			} else {
 				klog.V(2).Infof("Not restarting dnsmasq (--restartDnsmasq=false)")
 			}
--- a/pkg/dnsmasq/nanny.go
+++ b/pkg/dnsmasq/nanny.go
@@ -24,6 +24,8 @@ import (
 	"net"
 	"os/exec"
 	"strings"
+	"syscall"
+	"time"
 
 	"k8s.io/dns/pkg/dns/config"
 	"k8s.io/klog/v2"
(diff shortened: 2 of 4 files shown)


DamianSawicki:
The failed test fails identically at HEAD (issue https://github.com/kubernetes/dns/issues/646).

aojea:
I think you have several changes in the same commit that makes it hard to review.

In addition, what is the motivation to make this configurable, I don't think we should allow people to use this feature if is not for debugging, and if that is the case, you can perfectly do it with a simple bash script

DamianSawicki:
Thank you very much for a quick reaction, Antonio!

> I think you have several changes in the same commit that makes it hard to review.

Yeah, I replace `nanny.X()` with `if err := nanny.X(); err != nil { log }` for `X in {Kill, Start}`, which I'm happy to move to a separate commit, but everything else belongs together: there is a functionality change with the corresponding test, and the new test case requires expanding the testing fixture.

> I don't think we should allow people to use this feature

You've mentioned running a bash script. Do I understand correctly that users can also run such a script sending SIGUSR1 to a dnsmasq instance? If so, why don't you want to allow them to use this feature?

> In addition, what is the motivation to make this configurable

We can hardcode some interval, but I thought it would be safer to have things configurable, even just for turning the feature on/off.

The general motivation is to improve dnsmasq observability and not to require users to run custom scripts (or reach out to their cloud providers) to gain some insights about their kubedns load. Does it make sense to you?

aojea:
I can understand if you expose these metrics with a metrics/prometheus endpoints, that allows you to set alerts and integrate with common monitoring stacks  ... but you still needs to evaluate the performance implications on sending a periodic signal to the process

bowei:
Does dnsmasq handle syscall interruption gracefully? e.g. on receipt of signal, the syscall is canceled with EINTR.

DamianSawicki:
@bowei

> Does dnsmasq handle syscall interruption gracefully? e.g. on receipt of signal, the syscall is canceled with EINTR.

Yes, dnsmasq mentions specific actions on `SIGUSR1`, `SIGUSR2`, and`SIGHUP` in its [linux man page](https://linux.die.net/man/8/dnsmasq) and it is a very mature piece of software, so I cannot imagine it handling signals incorrectly. Just to be 100% sure, I've just checked the codebase, and there are many of instances of `errno == EINTR` and `errno != EINTR` throughout.

@aojea 

> you still needs to evaluate the performance implications on sending a periodic signal to the process

The statistics logged by dnsmasq on `SIGUSR1` contain not just the current utilisation, but also the maximal utilisation since the last `SIGUSR1`, so I was thinking about using `~5 min` intervals in practice (`1 min` at least, while `1 h` would still be better than nothing). The function `dump_cache` handling `SIGUSR1` is cheap (unless dnsmasq runs in debug mode when it does a full cache dump). With such infrequent and minimal additional load and correct handling of `EINTR`, performance implications should be negligible.

> I can understand if you expose these metrics with a metrics/prometheus endpoints, that allows you to set alerts and integrate with common monitoring stacks

For what you describe, the first step is still the current PR, which makes dnsmasq print statistics to logs, which we can later parse and expose with metrics/prometheus endpoints or set up log-based metrics. Do you think the current PR is only sensible if extended with these follow-up steps?

aojea:
I don't know, it really feels awkward running a process in production that you are sigkilling periodically ... but I don't know if this is also used elsewhere

DamianSawicki:
> that you are sigkilling periodically

For anyone reading this without much context, Antonio meant *sending a signal periodically*. The signal `SUGUSR1` does not kill the process, it's just a way of communicating with it. Dnsmasq is a rather old piece of software, and that's one way of communication that it supports, but I agree that it may feel awkward.

In any case, PRs to this repo are very rare, which means the resources are limited (two PRs with vulnerability fixes have been waiting for attention since July). The present PR (which actually was on my to-do list for months) is a rather low-hanging fruit to improve kubedns observability which is among priorities to help with supportability. Let's not rush and see what others have to say.

aojea:
problems on typing fast, thanks for clarifying, I meant what Damian said 😅

pwdtr:
+1, Good work, Damian!

bowei:
If we can do a test and include the results in the commit message to demonstrate that sending the signal is benign:

* Send constant stream of dnsperf requests to dnsmasq
* Enable the periodic interval of SIGUSR1

DamianSawicki:
/remove-lifecycle stale

I need a bit of free time to run the benchmark suggested by Bowei.

DamianSawicki:
> If we can do a test and include the results in the commit message to demonstrate that sending the signal is benign:
> 
> * Send constant stream of dnsperf requests to dnsmasq
> * Enable the periodic interval of SIGUSR1

I benchmarked the change with SIGUSR1 sent as frequently as every second for both UDP and TCP (with up to 50 concurrent TCP connections) and observed no negative performance impact. Added details to commit description.

Please review 🙏

DamianSawicki:
This is a (+70, -3) change, but since Antonio mentioned above that it includes several logical changes, I split it into 3 commits to further facilitate review:
 * (+6, -3) [clean-up](https://github.com/kubernetes/dns/pull/644/commits/41709441308064ccbcf5416b8fd524ffb947564e),
 * (+28, 0) [main change](https://github.com/kubernetes/dns/pull/644/commits/7bbed083f04c98147f705c9823ddda396d4e3ddc),
 * (+36, 0) [test](https://github.com/kubernetes/dns/pull/644/commits/9e83aaf4529340b96efce293a0c263690f2874e4).

DamianSawicki:
Hey @aojea @bowei @MrHohn! The PR is now ready for the final approval (and review perhaps, although there are no code changes compared to October: just rebased, added benchmarking results to a commit message, and spit the original commit in 3 parts), but since the PR has been inactive for almost 5 months, I'm pinging you so that it's not missed.

cmd/dnsmasq-nanny/main.go
```diff
@@ -69,6 +70,9 @@ Any arguments given after "--" will be passed directly to dnsmasq itself.
 		"interval to check for configuration updates")
 	flag.StringVar(&opts.kubednsServer, "kubednsServer", opts.kubednsServer,
 		"local kubedns instance address for non-IP name resolution")
+	flag.DurationVar(&opts.LogInterval, "logInterval",
+		opts.LogInterval,
+		"interval to send SIGUSR1 to dnsmasq which triggers statistics logging")
```
Michcioperz, line 75: might be useful to document 0 as disabling value

Michcioperz (approved):
Reviewing from my perspective as agreed. LGTM with one nitpick. This would be useful for supportability.

PR #647, opened 2024-10-14
Fix formatting in specification.md

limo520:


kundan2707 (commented):
@limo520 it is very trivial change. Can you check if any more change required in doc

Issue #655, opened 2024-11-20
Node local dns for dual-stack


DockToFuture:
We set up kubernetes dual-stack clusters and want to use node-local dns. When checking the code I see that only one IP familiy is supported at time (https://github.com/kubernetes/dns/blob/fe97ec9fa1ee596315827138f693b68bcab51ff8/cmd/node-cache/main.go#L104-L106). What is the reason for that? Can we also add dual-stack support to the kubernetes node-cache?

DockToFuture:
As an implementation proposal I suggest: https://github.com/kubernetes/dns/pull/657.

Issue #642, opened 2024-08-08
`node-local-dns` should support graceful shutdown with the `ready` plugin


jsws:
We are using `node-local-dns` with a CiliumLocal Redirect Policy as described [here](https://docs.cilium.io/en/stable/network/kubernetes/local-redirect-policy/#node-local-dns-cache) so we need `node-local-dns` to be able to become Not Ready once it receives a `SIGTERM` so time is given for the Endpoint to stop receiving traffic and for requests to be directed to `kube-dns` instead, before the pod is fully terminated. This will allow graceful rollouts of the `node-local-dns` daemonset. 

In https://github.com/kubernetes/dns/issues/485 it was requested the `ready` plugin be added to `node-local-dns` but it was closed on the basis it wouldn't serve any purpose as it would behave the same way as the `health` plugin. However I believe that's not the case and the `ready` plugin would allow `node-local-dns` to gracefully shutdown, since once it receives a `SIGTERM` the [`/ready` endpoint would start returning 503](https://github.com/coredns/coredns/blob/e5945d2767ee45b1a2a4a2aba8daf8c470c4519c/plugin/ready/ready.go#L48-L52) which can be used in a Readiness Probe on the pod to determine the pod is shutting down and give time for the Endpoint to be removed when used with the [`health` plugins `lameduck`](https://coredns.io/plugins/health/) setting. 
So I am again requesting the `ready` plugin be added to the plugins list [here](https://github.com/kubernetes/dns/blob/master/cmd/node-cache/main.go#L31).

LeoQuote:
Hi, I’m currently exploring this option( cilium + nodelocal dns), but I see no reason to use image provided by this project instead of using coredns directly, did you try to use coredns directly? Is there any hidden disadvantages?

I tried to use coredns instead and even drafted a PR for this. Hope that would help you https://github.com/coredns/helm/pull/173

kevin2000141:
Hi, I'm creating [this PR](https://github.com/kubernetes/dns/pull/669) to solve the readiness issue.

PR #657, opened 2024-11-20
Add dual-stack support for node-cache

DockToFuture:
Add dual-stack support for node-cache


DamianSawicki:
Implementation-wise, it would be ideal to have a test for this new feature.

Project-wise, let's get an agreement on https://github.com/kubernetes/dns/issues/642 before merging this. I've just reopened it so that the discussion can continue.

DamianSawicki:
> @DamianSawicki How is #642, which is about graceful shutdown/readiness, related to supporting dual-stack, i.e. IPv4 and IPv6? As of now `node-local-dns` only supports exclusively either IPv4 or IPv6, but not both at the same time. To me, these two features seem to be rather orthogonal to each other.
> 
> Please advise.

I'm terribly sorry, I somehow confused PR https://github.com/kubernetes/dns/pull/669, which is related to Issue https://github.com/kubernetes/dns/issues/642, with the present PR. Please disregard my comment.

ScheererJ:
No problem at all. Thanks for clarifying.

DamianSawicki (commented):
https://github.com/kubernetes/dns/issues/655 is closed, should this PR remain open?

If so, it would be good guard this change with a feature flag and add some tests.

DockToFuture:
@DamianSawicki is there some interest in dual-stack support for node-cache, as there were no reactions for a long time? Then I will adapt the PR.

DockToFuture:
Any opinion on that @Michcioperz @marqc?

axel7born:
I see an additional motivation for merging this PR:

Without dual-stack support in node-local DNS cache, some DNS queries will bypass the cache and hit CoreDNS directly due to resolver behavior differences.

Example with nslookup:
1. `nslookup` sets the Recursion Desired (RD) flag in DNS queries
2. CoreDNS resolves the query successfully but doesn't set the Recursion Available (RA) bit in responses
3. `nslookup` interprets the missing RA bit as an error and fails over to the next nameserver
4. The failover queries the IPv6 address of the kube-dns service
5. Since node-local DNS cache currently only intercepts IPv4, these IPv6 queries bypass the cache entirely and reach CoreDNS directly

This defeats the purpose of the node-local DNS cache for dual-stack clusters and negates its performance and reliability benefits for tools like nslookup and potentially other resolvers with similar behavior.

DockToFuture:
PTAL @DamianSawicki @Michcioperz @marqc

DockToFuture:
Can we get the dual-stack support in @DamianSawicki @Michcioperz @marqc @kl52752 @MrHohn
MrHohn or do you see any objections?

Michcioperz:
I am no longer engaged in DNS maintenance but I would like to point out that node-cache was moved out to https://github.com/kubernetes-sigs/node-local-dns

DockToFuture:
Thank you for the answer @Michcioperz I will create a new issue there and open a new PR.

PR #659, opened 2024-11-26
chore: bump go version

marqc:


pwdtr:
Hey, shouldn't we also update the version in https://github.com/kubernetes/dns/blob/master/Makefile#L43?

marqc:
> Hey, shouldn't we also update the version in https://github.com/kubernetes/dns/blob/master/Makefile#L43?

I believe this is the exact line I changed.

pwdtr:
> I believe this is the exact line I changed.

Sorry, my bad. It was just that go.mod has a major version pinned, while the image has patch version pinned. I was referring to this comment.

https://github.com/kubernetes/dns/commit/6a92f4ba4d54a1b683cc59957213f8e1a268d720#commitcomment-144374779

pwdtr:
Okay nevermind, 1.22 is 1.22.0 not the latest in this case I think.

heshamelsherif97:
Can we get this merged?

PR #664, opened 2024-12-16
docs: Presubmit triggering and grid validation in README#Release process

DamianSawicki:
This adds to the release instructions in `README.md` that a presubmit
should be triggered before bumping kubedns or nodecache (`cluster/addons/dns/`) versions in
kubernetes/kubernetes and the test grid https://testgrid.k8s.io/sig-network-gce#gci-gce-serial-kube-dns (and other related tabs) should be verified afterwards. This should prevent issues like https://github.com/kubernetes/kubernetes/pull/123590 and https://github.com/kubernetes/kubernetes/issues/129230 from happening in the future and/or speed up their detection.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,8 +66,10 @@ Follow these steps to make changes and release a new binary.
    The SHAs can be obtained by running the command `python parse-image-sha.py <TAG>`
    This will return the SHAs for kube-dns as well as node-cache images. Node-cache images are always promoted, kube-dns images are promoted if there is a change to kubedns/vulnerability fix.
 6. Images will be available in the repo registry.k8s.io/dns/. The node-cache image with tag 1.15.14 can be found at registry.k8s.io/dns/k8s-dns-node-cache:1.15.14. Older versions are at registry.k8s.io/k8s-dns-node-cache:<TAG>
-7. Submit a PR for the kubernetes/kubernetes repository to switch to the new
-   version of the containers. Example - https://github.com/kubernetes/kubernetes/pull/106189
+7. Prepare a PR for the kubernetes/kubernetes repository to switch to the new
+   version of the containers. Example - https://github.com/kubernetes/kubernetes/pull/106189.
+   Trigger the optional [presubmit](https://github.com/kubernetes/test-infra/pull/33962) `pull-kubernetes-e2e-gci-gce-kube-dns-nodecache` and correct your PR if needed before merging.
+8. Verify the kubedns-related and nodecache-related tabs of the test grid at https://testgrid.k8s.io/sig-network-gce for regressions caused by the new image and revert if needed.
    
 ## Version compatibility
 


README.md
```diff
@@ -66,7 +66,8 @@ Follow these steps to make changes and release a new binary.
    The SHAs can be obtained by running the command `python parse-image-sha.py <TAG>`
    This will return the SHAs for kube-dns as well as node-cache images. Node-cache images are always promoted, kube-dns images are promoted if there is a change to kubedns/vulnerability fix.
 6. Images will be available in the repo registry.k8s.io/dns/. The node-cache image with tag 1.15.14 can be found at registry.k8s.io/dns/k8s-dns-node-cache:1.15.14. Older versions are at registry.k8s.io/k8s-dns-node-cache:<TAG>
-7. Submit a PR for the kubernetes/kubernetes repository to switch to the new
+7. Verify the test grid at https://testgrid.k8s.io/sig-network-gce#gci-gce-serial-kube-dns for regressions caused by the new image.
```
aojea, line 69: This is a chicken an egg problem, you need to test it before submitting it, we can add a presubmit job optional that we can trigger on demand for this https://github.com/kubernetes/test-infra/blob/6cbed8179723b665fd8ad8a8af9a986d4a9f4596/config/jobs/kubernetes/sig-network/sig-network-misc.yaml#L790,

DamianSawicki: I see, so we couldn't detect it before merging with the current tooling. I'll reword the instruction then.

aojea:
/lgtm
/approve

as a follow up we may want to inrtoduce the optional presubmit so we don't risk to have to revert after merge

README.md
DamianSawicki: Done.

README.md
```diff
@@ -68,6 +68,7 @@ Follow these steps to make changes and release a new binary.
 6. Images will be available in the repo registry.k8s.io/dns/. The node-cache image with tag 1.15.14 can be found at registry.k8s.io/dns/k8s-dns-node-cache:1.15.14. Older versions are at registry.k8s.io/k8s-dns-node-cache:<TAG>
 7. Submit a PR for the kubernetes/kubernetes repository to switch to the new
    version of the containers. Example - https://github.com/kubernetes/kubernetes/pull/106189
```
marqc, line 70: nit: Can we test the new descriptors in the empty kind cluster as part of this point (`kubectl apply -f ...`)? To verify they are working and no other changes apart from image version bumps are necessary in the descriptors.

DamianSawicki: Wouldn't this manual test be equivalent to the new presubmits [suggested](https://github.com/kubernetes/dns/pull/664#issuecomment-2545524298) by Antonio? If so, then the automated approach feels much less-effort and more fail-proof to me.

marqc: ack. I wasn't aware of the automated presubmits PR. I agree that it's much better. Thanks.

marqc (approved):

marqc (approved):

PR #668, opened 2025-01-02
Bump golang.org/x/net to v0.33.0

marqc:
Fixes CVE-2024-45338


marqc:
This should be bumped by #661. If that PR turns out to be inferasible to merge, we can resuurect this PR.

PR #675, opened 2025-02-14
update OWNERS file

kl52752:
remove kl52752 from Owners file and add DamianSawicki instead.


DamianSawicki:
/lgtm
/approve

Code changes look ok; it would be ideal to add more details (who gets removed, who gets added) to commit description, PR description, or perhaps both.

PR #677, opened 2025-02-19
Bump non k8s feb 2025

dereknola:
## Changes
- Bumped various dependencies to latest available version that used go 1.22 (a move to 1.23 is planned on a future PR).
- Moved the weird third-party copy of the old skyDNS (I really don't want to know why this happened) off the old `coreos/etcd` import to the modern `etcd-io/etcd` import. You can see how this now aligns all usage of etcd around a single version `3.5.18`, not this weird old `v3.3.13+incompatible`.


dereknola:
@damdo can we just get rid of the `go vendor`? For a small codebase like this, it just balloons any PR and this feels very "old golang".

dereknola:
@DamianSawicki  can we just get rid of the `go vendor`? For a small codebase like this, it just balloons any PR and this feels very "old golang".

DamianSawicki:
> * Moved the weird third-party copy of the old skyDNS (I really don't want to know why this happened) off the old `coreos/etcd` import to the modern `etcd-io/etcd` import. You can see how this now aligns all usage of etcd around a single version `3.5.18`, not this weird old `v3.3.13+incompatible`.

I see https://github.com/coreos/etcd redirects to https://github.com/etcd-io/etcd, so it sounds good.

> @DamianSawicki can we just get rid of the `go vendor`? For a small codebase like this, it just balloons any PR and this feels very "old golang".

Interesting suggestion! If anything, it definitely should be done in a separate PR. I'm not the best person to answer this question, but I see Kubernetes uses `vendor` directory for some old repositories like https://github.com/kubernetes/ingress-gce, but not for new ones like https://github.com/kubernetes-sigs/nat64, so that would suggest there are no good reasons to keep `vendor`. WDYT @aojea?

/ok-to-test

DamianSawicki:
```
# k8s.io/dns/third_party/forked/skydns
third_party/forked/skydns/main.go:323:3: unknown field CAFile in struct literal of type "go.etcd.io/etcd/client/pkg/v3/transport".TLSInfo
make: *** [rules.mk:104: bin/amd64/e2e] Error 1
```
It seems that replacing 	`github.com/coreos/etcd/pkg/transport` with `go.etcd.io/etcd/client/pkg/v3/transport` in `third_party/forked/skydns/main.go` requires some adjustments.

DamianSawicki (changes requested):
Looking at https://github.com/etcd-io/etcd/pull/9470, it seems that the renaming [Move from CAFile to TrustedCAFile](https://github.com/kubernetes/dns/pull/677/commits/bd18ffa4f7e2db1c88b203505a6087adfa651659) is indeed all we need.

Personally, I'd prefer the commits `Move old coreos etcd module, use modern etcd-io based one` and `Move from CAFile to TrustedCAFile` to be squashed together and the respective parts of `go vendor` to be squashed into that commit and into `Bump non k8s dependencies`. This way, the repository shouldn't be broken at any point, even between merge commits. But perhaps `go vendor` can be omitted altogether if we decide to get rid of the `vendor` directory.

dereknola:
Okay, I've rebased this PR into two independent commits, each which work standalone, i.e. the `go vendor` has been done at each respective phase.

PR #679, opened 2025-02-24
Remove go vendor from project

dereknola:
Remove `vendor` folder as discussed in https://github.com/kubernetes/dns/pull/677. Ran `go mod tidy` afterwards.


DamianSawicki:
> Remove `vendor` folder as discussed in #677. Ran `go mod tidy` afterwards.

Actually, running `go mod tidy` on current HEAD also removes these two lines from `go.sum`, so it probably should have been run on the last merged PR. So, ideally, please isolate this change to a separate commit (ordered before the commit removing `vendor`).

/ok-to-test

dereknola:
@DamianSawicki Looks like the testing infra assumes that has to be a vendor.... do you know where this testing lives so we can change that?

DamianSawicki:
Apparently github does not support well changes involving close do 2m lines,
Screenshot 2025-02-27 at 17 34 24
 but I checked it manually, and the latter commit does what is expected: deletes the entire `vendor` and changes one line in `presubmits.sh`.
 
 /lgtm
 
As I already mentioned https://github.com/kubernetes/dns/pull/677#issuecomment-2671300529, I'd like someone with a broader context to approve it.

/assign @aojea

DamianSawicki (commented):
I should have left my previous comment via this "review" button.

aojea:
> Apparently github does not support well changes involving close do 2m lines, Screenshot 2025-02-27 at 17 34 24 but I checked it manually, and the latter commit does what is expected: deletes the entire `vendor` and changes one line in `presubmits.sh`. Thanks for taking care of this @dereknola!
> 
> /lgtm
> 
> As I already mentioned [#677 (comment)](https://github.com/kubernetes/dns/pull/677#issuecomment-2671300529), I'd like someone with a broader context to approve it.
> 
> /assign @aojea

/hold

sorry, didn't catch this, we use to keep the vendor folder so we can track and review the dependencies easily and also it allows to develop offline ... is there any special reason to remove it?

/cc @BenTheElder

DamianSawicki:
> is there any special reason to remove it?

There is currently not much active development here except for dependency bumps, which creates the impression that `vendor` "just balloons any PR". I'm not aware of any reasons other than those expressed by Derek in https://github.com/kubernetes/dns/pull/677#issuecomment-2669433353.

bowei:
If anything, I would love to see attempts to trim down the dependencies themselves. vendor/ is easily verified -- download the patch, apply it to your source tree and rerun `go mod tidy` to review if the code matches. I think it is also checked as part of the CI, so the vendor/ directory changes should not need to be examined.

DamianSawicki:
> download the patch, apply it to your source tree and rerun go mod tidy to review if the code matches.

Yeah, that's absolutely true, I don't think people review `vendor` in the same way as the rest of the code.

> I think it is also checked as part of the CI, so the vendor/ directory changes should not need to be examined.

Actually, it turned out in the present PR (see my https://github.com/kubernetes/dns/pull/679#issuecomment-2679534359 above) that this is currently not verified.

dereknola:
> If anything, I would love to see attempts to trim down the dependencies themselves. vendor/ is easily verified -- download the patch, apply it to your source tree and rerun `go mod tidy` to review if the code matches. I think it is also checked as part of the CI, so the vendor/ directory changes should not need to be examined.

I am not confident enough in this code base to trim down dependencies. Looking at the list, at face value this seems fairly light as is. Its mainly just DNS, etcd, k8s.io utility, and testing dependencies. 

All the direct `k8s.io` dependencies that this project uses don't have `vendor` anymore. In fact, it appears that every single direct dependency this project uses doesn't have vendoring anymore. There is clearly a push across the organization to move away from this older system. Cleaner repos beats out the ability to build offline.

DamianSawicki:
> All the direct `k8s.io` dependencies that this project uses don't have `vendor` anymore.

Interesting! With the notable exception of `kubernetes/kubernetes`, you are right!

bowei:
I'm ok with removing vendor/ it if there are no additional concerns from @aojea

BenTheElder:
> Yeah, that's absolutely true, I don't think people review vendor in the same way as the rest of the code.

We do review vendor heavily in kubernetes/kubernetes[^1] but I agree that most projects are not actually reviewing it.

> All the direct k8s.io dependencies that this project uses don't have vendor anymore. 

vendor doesn't make sense for library packages, because consumers of the library may have a different version of transitive dependencies, I assume nearly all of your k8s.io dependencies are just libraries. vendor is for binaries.

This likely isn't a meaningful datapoint.

> There is clearly a push across the organization to move away from this older system. Cleaner repos beats out the ability to build offline.

"Cleaner" is subjective, it's a directory you can easily ignore, but it does introduce some slight additional friction for developers.

In https://sigs.k8s.io/kind I dropped it early because `go get` / `go install` was a core part of the experience for the small main binary, but for anything else modern git has good options to clone a repo without fetching all of the historical blobs (e.g. `filter=blob:none`)

But we will likely never drop it for Kubernetes because it prevents us from dependency deletion even online (note that the go module proxy makes no availability guarantees for content, it is just a cache) and makes it easier to review dependency changes.

[^1]: I am one of the reviewers, we follow some guidelines under SIG Architecture and actually review the diff for anything problematic and/or fishy https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/vendor.md#reviewing-and-approving-dependency-changes

DamianSawicki:
Given the opinions from @aojea and @BenTheElder, I guess we can wait a bit for opposite opinions and close the PR. Thanks everyone for your detailed comments!

PR #685, opened 2025-04-19
Update distroless static-debian12 image

khaleksa:
Update Debian image:
-  from 12.8 (gcr.io/distroless/static-debian12@sha256:f4a57e8ffd7ba407bdd0eb315bb54ef1f21a2100a7f032e9102e4da34fe7c196) 
- to 12.10 (gcr.io/distroless/static-debian12@sha256:765ef30aff979959710073e7ba3b163d479a285d7d96d0020fca8c1501de48cb) version

Debian 12.10 ([release notes](https://www.debian.org/News/2025/20250315)) includes security fixes for CVE-2025-0395 and several others.

PR #688, opened 2025-05-15
Bump base images

Michcioperz:
followup to #685

distroless-iptables contains the fix for CVE-2025-0395 in node-cache image

base-debian12 contains the fix for CVE-2025-0395 in dnsmasq-nanny image

At this point I'm afraid to ask what's the difference between base-debian and debian-base, but it seems like bumping the amd64 COMPILE_IMAGE is also the right thing to do


DamianSawicki (approved):
Thanks for helping keep kubernetes/dns vulnerability-free!

/lgtm

Issue #453, opened 2021-06-04
NodeLocal DNS container hung on SIGTERM


rtheis:
We are still hitting the same problem reported by https://github.com/kubernetes/dns/issues/394.  The test failure occurred on Kubernetes version 1.21 with NodeLocal DNS cache version 1.17.3.

To recap, NodeLocal DNS container occasionally hangs on termination causing Kubernetes to kill the container using SIGTERM after the grace period has expired.  This leaves left over iptables rules on the node thus breaking DNS resolution.  Our theory is that there is iptables lock contention between NodeLocal DNS, Calico and/or Kubernetes.

prameshj:
Doesn't nodelocaldns startup right away though? The DNS downtime should be O(seconds). Is that what you observe?

It is possible for nodelocaldns to run into lock contention, but that usually has a log message. Anything in the logs?

rtheis:
@prameshj I was unable to collect any valuable logs at the time of the latest failure.  I assume there is some type of lock contention that causes the pod to hang on termination.

NodeLocal DNS does startup right away, but our test failure comes when we verify that DNS works after disabling NodeLocal DNS.  If we restart NodeLocal DNS then stop it again, that usually fixes the node.

prameshj:
> NodeLocal DNS does startup right away, but our test failure comes when we verify that DNS works after disabling NodeLocal DNS. 

Ah I see. Just to confirm - 1) test disables nodelocaldns 2) nodelocaldns pod is stuck handling sigterm and is killed by kubelet with iptables rules getting left over. 3) test times out with DNS failure?

Do you see a log line of the sigterm being handled - https://github.com/kubernetes/dns/blob/3b17e06879a46b2ec5d97105c611a315897fdb48/vendor/github.com/coredns/caddy/sigtrap_posix.go#L42 ?

It should call teardown in that case - https://github.com/kubernetes/dns/blob/3b17e06879a46b2ec5d97105c611a315897fdb48/cmd/node-cache/main.go#L53

It is possible that the pod handles sigterm and tries cleaning up iptables, but cannot get the lock. We do expose a metric on port 9353 for nodelocaldns lock errors, but we do not increment it for delete errors. We should check errors in https://github.com/kubernetes/dns/blob/3b17e06879a46b2ec5d97105c611a315897fdb48/cmd/node-cache/app/cache_app.go#L157 and update the metric.

rtheis:
[...]

PR #689, opened 2025-05-19
Add option for node-cache to reload in-process CoreDNS via a signal.

Michcioperz:
When the nodelocaldns addon is applied during a cluster upgrade, there is a risk of deadlock. Consider the following run:

1. node-local-dns ConfigMap is updated.
2. node-cache binary receives the update, writes a new Corefile.
3. CoreDNS's reload plugin picks up the new Corefile and begins to set up a new server instance.
4. node-local-dns DaemonSet is updated.
5. node-local-dns Pod receives SIGTERM with 30s grace period. It begins executing shutdown callbacks, locking instancesMu. One of the callbacks is sending to reload plugin's unbuffered `exit` channel to make its goroutine exit.
6. reload plugin finishes setting up a new server instance, and tries to Stop() the original instance. This requires the instancesMu mutex. instancesMu is locked and to unlock it, reload plugin would have to return to its main loop.
7. Grace period is exceeded and Pod is terminated without tearing down iptables rules.

This seems like a plausible root cause of https://github.com/kubernetes/dns/issues/453.

Reload plugin is still supported after this change. The addon continues to work as-is, but it provides a flag-gated way to reload via SIGUSR1.

In order to enable this new behavior, remove `reload` directives from the Corefile template and add `-reloadwithsignal` to container args.

Reload with SIGUSR1 doesn't cause the deadlock, because signal processing in coredns/caddy is sequential. If SIGTERM comes during SIGUSR1 processing, it will just wait (and hopefully reload and termination will fit in 30 seconds). IF SIGUSR1 comes during SIGTERM processing, well, that's too late for it.

### DIFF
--- a/cmd/node-cache/app/cache_app.go
+++ b/cmd/node-cache/app/cache_app.go
@@ -14,6 +14,7 @@ limitations under the License.
 package app
 
 import (
+	"fmt"
 	"net"
 	"os"
 	"strings"
@@ -52,6 +53,7 @@ type ConfigParams struct {
 	HealthPort           string        // port for the healthcheck
 	SetupIptables        bool
 	SkipTeardown         bool // Indicates whether the iptables rules and interface should be torn down
+	ReloadWithSignal     bool // Indicates config reload should be triggered with SIGUSR1, rather than expecting CoreDNS's reload plugin
 }
 
 type iptablesRule struct {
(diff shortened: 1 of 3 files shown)


Michcioperz:
@johnbelamaric Thanks for the ok-to-test.

I was hoping to raise this issue in CoreDNS, but I didn't have time to verify that this repro works on CoreDNS directly (though I don't see why it shouldn't). I'll try to find a moment tomorrow.

Personally I think reloading with signals is fine here, as we treat CoreDNS as a kind of opaque-box (we invoke its main() function in a goroutine). But I would be surprised if CoreDNS had to resort to using signals within itself, surely this can be made to make sense with mutexes? I will include your suggestion in the bug when I file it though.

DamianSawicki:
I see Bowei's questions are from May and they were probably sent this week by mistake, and this week Bowei wrote /approve twice, so the intention seems clear and I think I can just proxy-lgtm.

/lgtm

PR #690, opened 2025-05-21
fix: reorder cleanup sequence in node-cache teardown

zooneon:
The previous implementation removed the dummy interface before cleaning up iptables rules, which could leave orphaned iptables rules referencing the removed interface. This fix ensures proper cleanup order by:

1. First removing all iptables rules
2. Then removing the dummy interface

This change prevents potential issues where iptables rules might reference a non-existent interface after teardown.

I found this in my testing and confirmed that it was indeed causing errors.
When deleting or updating, the dummy interface is deleted first, causing a disconnect and DNS queries to fail.
I changed the order of the teardown logic and it worked without errors.
This change prevents disconnects from causing errors when updating.

### DIFF
--- a/cmd/node-cache/app/cache_app.go
+++ b/cmd/node-cache/app/cache_app.go
@@ -176,9 +176,6 @@ func (c *CacheApp) TeardownNetworking() error {
 		c.exitChan <- struct{}{}
 	}
 	var err error
-	if c.params.SetupInterface {
-		err = c.netifHandle.RemoveDummyDevice(c.params.InterfaceName)
-	}
 	if c.params.SetupIptables {
 		for _, rule := range c.iptablesRules {
 			exists := true
@@ -198,6 +195,9 @@ func (c *CacheApp) TeardownNetworking() error {
 			// Delete the rule one last time since EnsureRule creates the rule if it doesn't exist
 			err = c.iptables.DeleteRule(rule.table, rule.chain, rule.args...)
 		}
+	}
+	if c.params.SetupInterface {
+		err = c.netifHandle.RemoveDummyDevice(c.params.InterfaceName)
 	}
 	return err
 }


Michcioperz:
Rules created in initIptables don't reference the dummy interface, only the IP addresses which it is being assigned.

Did you run into an issue which is fixed by this PR? Can you talk more about it?

zooneon:
@Michcioperz , ah i'm sorry that my explanation wasn't clear enough.
As you said initIptables do not reference the dummy interface, but because of that rules that initIptables generate, DNS query packets are forwarding to "Local DNS" not "Cluster DNS", and in that process dummy interface is used.
So when dummy interface is deleted and iptable rules remain, the packets can not find the interface to the local DNS and are forced to error out.
As mentioned in the description, i discovered this while testing and actually saw the error occur, and when i changed the order of tearDown logic as i changed, it worked without error.

How i tested
- Query the NGINX pod that executed in the different namespace using the dig command every 0.1 seconds on the test pod
- Enter to the worker node where the test pod is running, check the iptable rules, dummy interface deletion order, and check if the test pod throws an error.

And finally, i think it makes sense to delete them in the reverse order in which they were created.

zooneon:
I'm attaching a video of my test.

delete dummy interface first, and after that delete iptable rules
current_version

delete iptable rules first, and after that delete dummy interface
modified_version

Each screen is described below.
- Left top : Run the watch command to watch 'iptables -t raw -nvL' every 0.5 seconds.
- Left bottom : Run the watch command to watch 'ip a' every 0.5 seconds.
- Right top : Command to delete node local dns.
- Right bottom : Test pod that quering to nginx pod that executed in the different namespace with dig command, to see the error being thrown.

Michcioperz:
Ohh, yeah, I think I understand now. If the interface is deleted first, but rules are still there, there will be a short moment where the traffic is still kept on the node, because it will bypass the Kubernetes Service handling iptables rules; at the same time there will be no interface on the node those packets can go to, so they will go into the void.

I was worried this could be related to an issue I'm fixing, and it is the same symptoms, but different amount of time and cause.

PR #693, opened 2025-05-22
Bump base-debian12 to fix CVE-2024-13176

Michcioperz:
This vulnerability was once revealed to me in ~~a dream~~ an internal check.

Change-Id: I6d155f82de563068ec6fabf52f4f93c75d2cd402


images/dnsmasq/Makefile
```diff
@@ -28,8 +28,8 @@ OUTPUT_DIR := _output/$(ARCH)
 export DOCKER_CLI_EXPERIMENTAL=enabled
 
 # Multiarch image
-# Uploaded: Mar 28, 2025
-BASEIMAGE ?= gcr.io/distroless/base-debian12@sha256:27769871031f67460f1545a52dfacead6d18a9f197db77110cfc649ca2a91f44
+# Uploaded: May 22, 2025
```
DamianSawicki, line 31: As far as I can see, the image is from May 19th.

DamianSawicki (changes requested):

images/dnsmasq/Makefile
DamianSawicki: Well, perhaps you typed today's date from the calendar 😅

PR #694, opened 2025-05-26
Enable optional TLS on nodecache metrics endpoint

### LINKED ISSUE #692: feature: TLS-support for node-local-dns metrics endpoint
Currently the `/metrics` endpoint served by node-local-dns does not include support for TLS.

Adding TLS would require at least configuring the location of cert-file and key-file. Additionally adding configuration options to things such as client CA file would make the feature more convenient to use.

At the moment the endpoint configuration contains only the address and port number where to serve the metrics. This is done via command line flag as with the rest of node local DNS configuration. Adding multiple new command line flags to configure TLS seems cumbersome. What is the community opinion on how this should be implemented, I have currently thought of
- command line flags
- add configuration file for TLS, whose location is given as command line flag.

Which one is preferred?

nuhakala:
Fixes #692. Chose to do the configuration with flags instead of extra config file because thought that it will be simpler and more in line with the project overall.

Adapted the implementation for CoreDNS metrics TLS support to nodecache. The original PR here https://github.com/coredns/coredns/pull/7255
- TLS support optional (backward compatible)
- TLS can be configured through command line flags, including the key files and client certificate.
- Added tests

### New command line flags
- `tls-cert-file` default value `/etc/ssl/tls.crt`
- `tls-private-key-file` default value `/etc/ssl/tls.key`
- `client-auth-type` default value `NoClientCert`
- `client-ca-file` default value empty
- `tls-min-version` default value TLS version 13
- `tls-enabled` default value false

In order to use the feature, one needs to give as flag `-tls-enabled=1`. The defaults are used and the existence of TLS key-files is checked and error is thrown if no files are found. The defaults can be changed by using the corresponding flags.


nuhakala:
@bowei @MrHohn Can I get a review on this so that the PR can move forward?

nuhakala:
@bowei @MrHohn Could I get review on this?

kashifest:
This PR is up since May 2025. Can we get some reviews on this @DamianSawicki.

PR #706, opened 2025-06-09
Pin k8s.io/kuberenetes + dependency bumps

dereknola:
https://github.com/kubernetes/dns/pull/704 is in a broken state and I lack permissions to fix it. Opening this PR instead.

## Changes 
- Pin `k8s.io/kubernetes` to match the rest of the k8s dependencies
- Keep `go 1.23.0` the above issue was causing go to be bumped when it should not be.

## Validation
`make` works locally

PR #709, opened 2025-06-16
Bump Coredns version to include PR https://github.com/coredns/coredns…

navnandys:
Bump Coredns version to include PR
https://github.com/coredns/coredns/pull/6999

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.23.0
 
 require (
 	github.com/coredns/caddy v1.1.1
-	github.com/coredns/coredns v1.11.3
+	github.com/coredns/coredns v1.12.1
 	github.com/coreos/go-systemd v0.0.0-20191104093116-d3cd4ed1dbcf
 	github.com/miekg/dns v1.1.63
 	github.com/onsi/ginkgo v1.16.5


DamianSawicki (changes requested):
Is the bump of `github.com/coredns/coredns` from `v1.11.3` to `v1.12.1` backwards compatible? Looking briefly at https://github.com/coredns/coredns/releases it seems so, but IIRC there have been some surprises in this respect in the past.

/cc @Michcioperz

DamianSawicki:
There's gonna be a problem with this bump because coredns v1.12.1 [requires](https://github.com/coredns/coredns/blob/v1.12.1/go.mod#L39-L41)
```
	k8s.io/api v0.32.3
	k8s.io/apimachinery v0.32.3
	k8s.io/client-go v0.32.3
```
and it's been the policy in kubernetes/dns to keep such k8s dependencies at the latest supported minor (currently `30` rather than `32`).

Unfortunately, this also [affects](https://github.com/kubernetes/dns/pull/710#issuecomment-3007792621) our automatic bumps. FYI @dereknola @Michcioperz

dereknola:
Also we never should worry about `coredns/coredns` usage of k8s dependencies, because they are only used in their `plugin/kubernetes` which is explicitly stated as a [**replacement** of this repo `kube-dns`.](https://github.com/coredns/coredns/tree/b2a2a5f6480ffc584411305277172aa7fa03cf1b/plugin/kubernetes#description)

DamianSawicki:
@navnandys Closing this PR because https://github.com/kubernetes/dns/pull/706 predates and covers it.

@dereknola I haven't been following up on https://github.com/kubernetes/dns/pull/706 because I've been hoping that we can make dependabot do the work for us (see e.g. https://github.com/kubernetes/dns/pull/710#issuecomment-3007792621).

PR #712, opened 2025-07-16
Bump distroless-iptables to v0.6.12

khaleksa:
Upgrade the distroless-iptables image to version 0.6.12, based on Debian 12.11, to address various CVEs including CVE-2024-13176. 
The complete list of security fixes in the Debian 12.11 release notes is available at: https://www.debian.org/News/2025/20250517


DamianSawicki:
> I didn't have to bump those myself, only IPT, so they should be still okay :P

Thanks for letting us know!

/ok-to-test

PR #756, opened 2026-03-05
CVE patch

jeffrhea-wf:
Updating base image and go libraries to resolve the following CVEs:
CVE-2026-22796
CVE-2026-22796
CVE-2025-15467
CVE-2025-68151
CVE-2025-64702

Notes:
Coredns version 1.14.2 brings in some other updated lib that causes the pull-kubernetes-dns-test test to fail.  Confirmed that only downgrading coredns doesn't resolve this failure, so I know this is a transitive dependency updated.  I have not done any more testing to confirm which transitive dependency is the issue at this time.


DamianSawicki (changes requested):
Speaking of CVEs, since you're already bumping github.com/coredns/coredns from `v1.13.2` to `v1.14.0`, it might be worth bumping to https://github.com/coredns/coredns/releases/tag/v1.14.1, which fixes some CVEs from `v1.14.0` according to the release description in the link.

IIRC (@Michcioperz to keep me honest), sometimes bumps of github.com/coredns/coredns introduce breaking changes, but this time I don't see anything suspicious in the release descriptions.

Other than that the PR looks fine. Thanks for taking this up!

/ok-to-test

Michcioperz:
I don't see anything particularly scary in the changelogs, just to point out there's even a 1.14.2 now

jeffrhea-wf:
Good call on the coredns changes.  Updated to latest 1.14.2.

DamianSawicki:
Thanks @jeffrhea-wf! Please take care of the CLA (https://github.com/kubernetes/dns/pull/756#issuecomment-4008279708) and we should be ready to merge the PR :)

pkg/e2e/dns/dns.go
```diff
@@ -61,23 +61,24 @@ func (kd *KubeDNS) Start(name string, args ...string) {
 		kd.isRunning = false
 	}()
 
-	// dns service
+	// SkyDNS listens as soon as Run() starts the skydns goroutine.
 	om.Eventually(func() error {
-		conn, err := net.Dial("tcp", "localhost:10053")
+		conn, err := net.Dial("tcp", "127.0.0.1:10053")
```
DamianSawicki: Updating timeouts is probably fine, but the need to replace `localhost` by `127.0.0.1` looks like a behavior change. Why is this need?

jeffrhea-wf: Sorry about that Damian, I had just been trying to test the pull-kubernetes-dns-test and confirm the results.  Change has been removed

PR #761, opened 2026-04-15
Configurable Prometheus endpoint via Corefile variable

### LINKED ISSUE #542: TCP listen on port 9253 on all interface addresses creates security vulnerability
K8s cluster is installed on VMs/worker nodes having multiple interfaces. Couple of interfaces have external connectivity towards DC-GW. dns-node-cache (node-local-dns) listens on 9253 for prometheus. But this TCP listens on all the interfaces including external interfaces.
This makes cluster vulnerable to DoS attack.

worker-pool1-z92h78yx-stack:~ # lsof -i -n -P | grep :9253
node-cach   4720     root    8u  IPv6    41286      0t0  TCP *:9253 (LISTEN)

Isn't there a way to configure specific address to bind ? Why can't just listen on nodes InternalIP ?

terror96:
Currently, the Prometheus metrics endpoint in the node-cache Corefile was hardcoded to listen on all interfaces (e.g. "prometheus :9253"). This means the metrics port is exposed on every network interface of the node, which is undesirable in security-sensitive environments where metrics scraping should be restricted to a specific network or IP address.

This change introduces a new Corefile template variable __PILLAR__PROMETHEUS__ENDPOINT__ that is substituted at runtime with the value provided via the new command-line argument `prometheus-listen-address` which defaults to ":9253". This allows operators to control which address the Prometheus endpoint binds to.

If the variable is not present in the Corefile, the behaviour is unchanged.

Fixes #542


terror96:
Ping @dpasiukevich

terror96:
Ping @bowei

terror96:
Closing this pull request in favor of https://github.com/kubernetes-sigs/node-local-dns/pull/16.

